PR #54, opened 2016-11-01
Correcting mostly small typos while I was reading the README

gae123:
You will have hard time to see what I changed because you mostly use a line per paragraph. I suggest you limit lines to less than 80 characters so they are easily side by side diffable and let the .md processor join them together.


bvandewalle:
Thanks a lot for the edits @gae123

PR #57, opened 2016-11-03
NEW: Activating all containers, even without IPs.

bvandewalle:
The goal here is that even if an IP is not directly visible through the Docker API/Event API (CNI for example), the container should be activated and it is the responsibility of the PolicyResolver to figure out the IP. This is typically used for Kubernetes Integration.

Detailed additions:
Adding IPs field into the PUPolicy
Validating the IP received back inside the PUPolicy
Using the PUPolicy IPs instead of the Runtime IPs.


bvandewalle:
retest this

PR #67, opened 2016-11-08
Added new testcases for iptables supervisor

amitlimaye:


supervisor/iptables_test.go
```diff
@@ -105,3 +107,103 @@ func TestSuperviseACLs(t *testing.T) {
...
+		Address:  "20.20.0.0/16",
+		Port:     "80",
+		Protocol: "tcp",
+	}}
+	containerInfo.Runtime.SetIPAddresses(map[string]string{"bridge": "30.30.30.30"})
+	containerInfo.Policy.PolicyIPs = []string{"30.30.30.30"}
+	m := s.ipt.(TestIptablesProvider)
+
+	m.MockNewChain(t,func(table,chain string)(error){
+		if(table == "raw" ){
```
bvandewalle: Why are you specifically checking if the Table is raw before returning an error ? 
If I understand your test ace correctly you want to emulate an error when calling NewChain, isn't ?

supervisor/iptables_test.go
```diff
@@ -105,3 +107,103 @@ func TestSuperviseACLs(t *testing.T) {
...
+		}
+		return nil
+	})
+	err = s.Supervise("12345",containerInfo)
+	
+	if(err == nil){
+		t.Errorf("ignored Error from addContainerChain:mangle")
+		t.SkipNow()
+	}
+	//Check if there are now rules anymore
```
bvandewalle: you mean you want to check if there are still  rules or no more rules ?

supervisor/iptables_test.go
```diff
@@ -105,3 +107,103 @@ func TestSuperviseACLs(t *testing.T) {
...
+	rules,err = m.ListChains("12345")
+	if( len(rules) != 0 ){
```
bvandewalle: Will always return nil, since ListChains is the default implementation of the mock

bvandewalle:
@amitlimaye It seems tests are not passing/ code is not building. Did you try to run ./.tests.sh locally first ? (on your linux machine :) )

supervisor/iptables_test.go
amitlimaye: THe same new chain function is called from multiple places in addComtaomer chain function for different tables. 
Without this check we will always fail on the first check and never excercise the other paths

supervisor/iptables_test.go
```diff
@@ -105,3 +107,103 @@ func TestSuperviseACLs(t *testing.T) {
...
+			return errors.New("Append Failed")
+		}
+	})
+	err := s.Supervise("12345",containerInfo)
+	if(err == nil){
+		t.Errorf("ignored Error from Append")
+		t.SkipNow()
+	}
+	//Check if there are now rules anymore
+	rules,err := m.ListChains("12345")
```
amitlimaye: Will implement listChains again to call the actual listchains here. 
The step is to check if we cleaned up right after getting an error

bvandewalle: But none of your methods are calling the actual IPTable implementation, and they shouldn't as those tests cover the logic in supervisor and not the logic in IPTables. if you want to check that rules are cleaned up as part of `Unsupervise` , you need to mock the DeleteChains etc methods to make sure that `Unsupervise` is going through those

amitlimaye: yep agreed will remove these checks and mock the delete method

PR #71, opened 2016-11-14
Fixed: use Logrus instead of glog

Dogild:


bvandewalle:
Fix Travis and your coverage please :)

PR #80, opened 2016-11-22
NEW: New supervisor (IPSet)

bvandewalle:
An Alternative implementation of the Supervisor by using IPSets. It should reduce complexity by an order of magnitude.


monitor/docker.go
```diff
@@ -334,23 +324,15 @@ func (d *dockerMonitor) syncContainers() error {
 
 		if err != nil {
 			log.WithFields(log.Fields{
-				"package":   "monitor",
-				"monitor":   d,
-				"error":     err,
-				"container": container,
+				"package": "monitor",
```
satyamsi: Shouldn't we have the container Ids which failed in the log ?

monitor/docker.go
```diff
@@ -334,23 +324,15 @@ func (d *dockerMonitor) syncContainers() error {
...
+				"error":   err,
 			}).Error("Error Syncing existing Container")
-
```
satyamsi: How do we deal with errors while syncing containers ?

monitor/docker.go
bvandewalle: Yes, but the issue is that in some cases the "Container" object could be nil. (not given by the Docker API).

monitor/docker.go
bvandewalle: When syncing, if an error happens (Example the PolicyResolver fails), we simply do not activate the container in Trireme. Depending on the default policy, this means that the container will/will not be the open Network Connections

monitor/docker.go
bvandewalle: We were actually logging the container.id in the past, but that could result in a Kernel panic if the container object is nil. Ideally this should never happen and be caught by the Docker event APi upstream ...

satyamsi: so it will print nil but in other cases we will see the id.

monitor/docker.go
satyamsi: ok

PR #98, opened 2016-12-09
Utils

satyamsi:


supervisor/ipsetsupervisor_test.go
```diff
@@ -13,9 +14,10 @@ func doNewIPSetSupervisor(t *testing.T) *ipsetSupervisor {
 	pe := mockenforcerDefaultFQConfig(t)
 	ipt := provider.NewTestIptablesProvider()
 	ips := provider.NewTestIpsetProvider()
+	ipu, _ := iptablesutils.NewIpsetUtils(ipt, ips)
```
Dogild: Don't ignore the error here

supervisor/ipsetsupervisor_test.go
```diff
@@ -29,33 +31,34 @@ func TestNewIPSetSupervisor(t *testing.T) {
...
```
Dogild: Same as above, don't ignore the error

PR #100, opened 2016-12-09
NEW: Adding a bash metadata extractor

bvandewalle:


dstiliadis:
Can we add an example of a bash extractor for documentation purposes?

bvandewalle:
@satyamsi : This is a `bashDockerMetadataExtractor` specifically Aimed for Docker Containers as the parameter to the function is of type Docker.

We also got the `defaultDockerMetadataExtractor` .
I think those names reflect the goal of each  extractor pretty clearly. How would you call them otherwise ?

bvandewalle:
@dstiliadis : Example now supports an external Metadata extractor with the -metadata flag.
There is also an extra binary in example that povides an example external extractor :)

PR #112, opened 2016-12-21
New: Support for Docker Swarm, VXLAN, MacVLAN, IPVLAN, etc using remote enforcers

dstiliadis:


dstiliadis:
Unit tests added in additional PR

cmd/remote_enforcer.go
```diff
@@ -248,10 +296,15 @@ func (s *Server) Supervise(req rpcwrapper.Request, resp *rpcwrapper.Response) er
...
 
 	runtime := policy.NewPURuntimeWithDefaults()
+
 	puInfo := policy.PUInfoFromPolicyAndRuntime(payload.ContextID, pupolicy, runtime)
-	puInfo.Runtime.SetPid(os.Getpid())
-	log.WithFields(log.Fields{"package": "remote_enforcer",
-		"method": "Supervise",
+
+	// TODO - Set PID to 1
+	puInfo.Runtime.SetPid(1)
```
satyamsi: We changed this from os.Getpid() to 1. This cant be correct. If we dont need this, it should be cleaned up/not called.

example/main.go
```diff
@@ -30,6 +36,57 @@ func init() {
 	flag.StringVar(externalMetadataFile, "m", "", "Description")
 }
 
+// swarmExtractor is an example metadata extractor for swarm that uses the service
```
satyamsi: This should be in example/common.go so agent/datapath/standalone can be used to test encryption without launching full blown agent.

PR #114, opened 2016-12-22
New: Deny Policies

dstiliadis:
Trireme supported a basic whitelist model where rules defined just accept actions. 

With this PR we are enhancing Trireme to accept higher priority deny rules. One can use this in 
several use cases: For example, a security team defines some policies that should never be violated
and prevent unauthorized action, while an application team defines the accept rules. An additional 
use case would allow an administrator to quarantine an application that is misbehaving.


dstiliadis:
Unit tests added in different PR. codecov can be ignored.

PR #115, opened 2016-12-27
Refactor of supervisor

dstiliadis:
Refactoring supervisor. Organize with iptables/ipsets as different packages. Added significant number of unit tests. 

The problem was that it was difficult to maintain and customize for some of the remote functions
since they are diverging from the local implementation. For example we can remove the raw tables in remote implementation. The refactor will allow us to simplify these changes.


configurator/configurator.go
```diff
@@ -92,45 +69,36 @@ func NewTriremeWithDockerMonitor(
...
+		trireme := trireme.NewTrireme(serverID, resolver, proxySupervise, proxyEnforce)
+		monitor := monitor.NewDockerMonitor(DefaultDockerSocketType, DefaultDockerSocket, trireme, dockerMetadataExtractor, eventCollector, syncAtStart)
+		return trireme, monitor, proxySupervise.(supervisor.Excluder)
 	}
 
-	enforcer := enforcer.NewDefaultDatapathEnforcer(serverID, eventCollector, nil, secrets)
-	IPTsupervisor, err := NewDefaultSupervisor(eventCollector, enforcer, networks)
-	// Make sure that the Supervisor was able to load. Panic if its not there.
+	localEnforcer := enforcer.NewDefaultDatapathEnforcer(serverID, eventCollector, nil, secrets, remoteEnforcer)
+	// localSupervisor, err := NewDefaultSupervisor(eventCollector, localEnforcer, networks)
```
satyamsi: merging problem maybe ? dont need this line.

configurator/configurator.go
```diff
@@ -107,7 +85,8 @@ func NewTriremeWithDockerMonitor(
 	}
 
 	localEnforcer := enforcer.NewDefaultDatapathEnforcer(serverID, eventCollector, nil, secrets, remoteEnforcer)
-	localSupervisor, err := NewDefaultSupervisor(eventCollector, localEnforcer, networks)
+	// localSupervisor, err := NewDefaultSupervisor(eventCollector, localEnforcer, networks)
```
satyamsi: Merge leftover maybe. Needs to be removed.

dstiliadis: It shows the two options with default supervisor or ipset based supervisor. Was left there for example reasons. I will add a comment.

PR #116, opened 2016-12-27
Fixing glide reference on example

dstiliadis:
After latest commit glide.lock in example needs to be updated.


primalmotion:
Glide must be updated through glide.yaml. Glide.lock should be removed from git.

PR #120, opened 2016-12-29
FEATURE: support for matching prefixes in labels

dstiliadis:
This feature allows Trireme to do longest prefix matching on label values. There are several 
use cases that allow a more flexible policy definition. Examples : 

Assume that an implementation has labels app=db1 app=db2 app=db3 . One can add a policy to 
allow traffic from app=db* 

Alternatively an implementation can use domain names as namespaces. For example 
domain=com.example.dev domain=com.example.qa domain=com.example.prod 

The a policy can be simple allow traffic from com.example.* 

The implementation uses the classic technique of maintaining hash tables for different prefix lengths.

### DIFF
--- a/enforcer/lookup/lookup.go
+++ b/enforcer/lookup/lookup.go
@@ -21,24 +21,24 @@ type intList []int
 //PolicyDB is the structure of a policy
 type PolicyDB struct {
 	// rules    []policy
-	numberOfPolicies int
-	equalPrefixes    map[string]intList
-	equalMapTable    map[string]map[string][]*ForwardingPolicy
-	notEqualPrefixes map[string]intList
-	notEqualMapTable map[string]map[string][]*ForwardingPolicy
-	notStarTable     map[string][]*ForwardingPolicy
+	numberOfPolicies       int
+	equalPrefixes          map[string]intList
+	equalMapTable          map[string]map[string][]*ForwardingPolicy
+	notEqualMapTable       map[string]map[string][]*ForwardingPolicy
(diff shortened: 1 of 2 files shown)


enforcer/lookup/lookup.go
```diff
@@ -29,14 +32,38 @@ func NewPolicyDB() (m *PolicyDB) {
 	m = &PolicyDB{
 		numberOfPolicies: 0,
 		equalMapTable:    map[string]map[string][]*ForwardingPolicy{},
+		equalPrefixes:    map[string][]int{},
```
satyamsi: New type for the []int{} will be better such as IntList. Which has an interface with a method called SortedInsert to start with.

enforcer/lookup/lookup.go
```diff
@@ -29,14 +32,38 @@ func NewPolicyDB() (m *PolicyDB) {
...
 		notEqualMapTable: map[string]map[string][]*ForwardingPolicy{},
-		starTable:        map[string][]*ForwardingPolicy{},
+		notEqualPrefixes: map[string][]int{},
 		notStarTable:     map[string][]*ForwardingPolicy{},
 	}
 
 	return m
 }
 
+func sortedInsert(array []int, value int) []int {
```
satyamsi: The new type say IntList should be used as a receiver for this method

enforcer/lookup/lookup.go
dstiliadis: It inserts on an ascending list though and we need a descending list.

PR #125, opened 2017-01-10
namespace switching code not built for mac

amitlimaye:
A small change to disable setns call on non linux systems. This allows unit tests to run there


amitlimaye:
setns builds disabled for non-linux systems.

PR #133, opened 2017-01-13
Remove dependency for Raw table from remote enforcer

dstiliadis:
This PR removes the dependency of the raw table at least from the remote enforcers. It is still needed to support source NAT in the case of docker bridge.


dstiliadis:
This PR has been absorbed with the Linux process support. Closing.

PR #135, opened 2017-01-13
Update README.md

ruvnet:
added google analytics tracking code to readme.


bvandewalle:
@ruvnet : can you also add it to kube-trireme ?

PR #136, opened 2017-01-21
Support for any Linux process

dstiliadis:
With this PR , Trireme is not just for containers any more, but it can support any arbitrary
Linux processes. It will map processes to cgroups and use the net_cls facilities to 
properly capture packets and enforce policy across any Linux process.


bvandewalle:
Let's try to increase the Code coverage before merge.

PR #149, opened 2017-02-01
Excluder interfaceimpl

amitlimaye:
Implementation of AddExcluder interface for remote/local enforcer. 
This allows you to add ip's to be excluded.


cmd/remoteenforcer/remoteenforcer.go
```diff
@@ -279,6 +280,8 @@ func (s *Server) InitSupervisor(req rpcwrapper.Request, resp *rpcwrapper.Respons
 			}
 			return err
 		}
+		s.Excluder = supervisorHdl
+		s.Supervisor = supervisorHdl
```
dstiliadis: can you change this supervisorHandler (more readable).

PR #151, opened 2017-02-03
Remote retries

amitlimaye:
Increased the number of retries before the client reports an error.
We are seeing the remote enforcer is taking a little more time the retries are now made configurable as an env variable so that we can configure it.


amitlimaye:
Retries is an env variable which can be tuned

PR #160, opened 2017-02-09
fixed: don't try to compile linux only thing on linux.

primalmotion:
This breaks all IDEs


primalmotion:
nope. that's not the issue

satyamsi:
I tried this there are some crazy things with cgo here. I have this working for netfilter_linux.go but the same technique doesnt work for this. 

Also, something that’s not obvious is that the build primitive in a go file has to be followed by a blank line otherwise its not a build primitive.

Satyam

PR #163, opened 2017-02-12
Trireme networks programmable through policy

dstiliadis:
Problem: 
The existing Trireme implementation requires the set of networks where
Trireme auth must be applied. This was statically configured. There are use cases where
we need to dynamically update the set of networks based.

PR:
With this PR the Trireme target networks are part of the policy interface. The set of networks
can be customized for each processing unit and there is no need to have the same list 
for all processing units. In addition, the set of Trireme networks can be updated dynamically 
with a policy update. 

Breaking API: This will break implementations that use Trireme since it changes the API.


supervisor/iptablesctrl/acls.go
```diff
@@ -64,7 +64,6 @@ func (i *Instance) chainRules(appChain string, netChain string, ip string) [][]s
 		i.appAckPacketIPTableContext,
 		i.appPacketIPTableSection,
 		"-s", ip,
-		"-p", "tcp",
```
satyamsi: We make it protocol agnostic for chainRules but not for cgroupChainRules. Will that be dealt with later or did we forget to update that ?

supervisor/ipsetctrl/ipsetctrl.go
```diff
@@ -193,7 +191,8 @@ func (i *Instance) addAllRules(version int, appSetPrefix, netSetPrefix string, a
 
 // Start implements the start of the interface
 func (i *Instance) Start() error {
-	if err := i.setupIpset(triremeSet, containerSet); err != nil {
+	// FIX THIS
```
satyamsi: Now, with this PR or later ?

PR #169, opened 2017-02-18
Fix race with remote enforcer FlowEntries.Len() cant be accessed w/o …

satyamsi:
…lock


dstiliadis:
This is not needed. Bigger PR completely refactors the flow collection and avoids the locks.

PR #174, opened 2017-02-27
Rpc secrets

amitlimaye:
Using random generated strings for hmac per remote enforcer. 
and a random key for stats channel


enforcer/proxy/enforcerproxy.go
```diff
@@ -200,25 +202,29 @@ func NewProxyEnforcer(mutualAuth bool,
 	rpchdl rpcwrapper.RPCClient,
 	cmdArg string,
 ) enforcer.PolicyEnforcer {
-
+	statsServersecret, err := crypto.GenerateRandomString(32)
+	for err != nil {
+		statsServersecret, err = crypto.GenerateRandomString(32)
+	}
```
dstiliadis: Why do we need this? Why not abort with an error ? Its very unlikely to fail.

PR #181, opened 2017-03-09
Fix: ExcludeWindows platform, because

bvandewalle:


satyamsi:
Bernard, check pull request 180. This fix is not needed.

satyamsi:
Guys, 

I have pr-180 for this. Can we approve that as that takes care of all files.

Satyam

PR #184, opened 2017-03-20
Fix to use sandboxkey in extractors where it is available and use it to switch namespaces 

amitlimaye:


bvandewalle:
As discussed, not needed anymore. Thanks for the PR though.

PR #194, opened 2017-04-03
Fixemptyip

dstiliadis:


bvandewalle:
@dstiliadis : Isn't that an issue that should also be solved in the Monitor/MetadataExtractor ? (Even though it is also good to add the extra check here for sure)

dstiliadis:
Yes, it should be handled there, but sometimes they can fail to do it and this creates a mess.

PR #198, opened 2017-04-12
Exit conditions

dstiliadis:
Clean up exit conditions for a graceful shutdown. Make sure that all go routines are properly shutdown.

### DIFF
--- a/supervisor/proxy/supervisorproxy.go
+++ b/supervisor/proxy/supervisorproxy.go
@@ -6,8 +6,6 @@ import (
 	"fmt"
 	"strconv"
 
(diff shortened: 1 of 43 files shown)


supervisor/proxy/supervisorproxy.go
```diff
@@ -169,24 +145,3 @@ func (s *ProxyInfo) InitRemoteSupervisor(contextID string, puInfo *policy.PUInfo
 	return nil
 
 }
```
bvandewalle: Why are we removing the call to AddExcludedIPs ?

PR #200, opened 2017-04-13
Remoteenforcer host

amitlimaye:
Code to ensure that we don't launch remote enforcer in host network namespace. 
Simple check to see if the container network namespace is not the same as init process.

### DIFF
--- a/processmon/processmon.go
+++ b/processmon/processmon.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"syscall"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/aporeto-inc/trireme/cache"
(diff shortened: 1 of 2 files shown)


processmon/processmon.go
```diff
@@ -178,6 +179,25 @@ func (p *ProcessMon) LaunchProcess(contextID string, refPid int, rpchdl rpcwrapp
...
+				"Method":  "LaunchProcess",
+				"Error":   "Refuse to launch an enforcer in the host namespace",
+			}).Info("Refused to Launch a remote enforcer in host namespace")
+			return nil
+		}
+	} else {
+
+		log.WithFields(log.Fields{
+			"package": "processmon",
+			"Error":   fmt.Sprintf("Hoststat err %v,PidStat err %v", hoststaterr, pidstaterr),
```
primalmotion: Can you use fields instead of sprintifing the errors?

primalmotion: *two* fields

PR #205, opened 2017-04-17
NEW: DockerMonitor No-Kill option

bvandewalle:
This PR adds an option on the DockerMonitor so that it can be spawned with a no-kill option in case the policy setup failed.

The way it works is in case Trireme returns an error for handling the start event of the container, the container isn't killed and continues to run. It is useful in environment such as Kubernetes where killing the containers might destroy the node completely


bvandewalle:
WIP. do not merge yet.

PR #206, opened 2017-04-19
Fixed: logs for debugs

satyamsi:


satyamsi:
not merging this. But found a major issue with the cache - we never delete and that causes all sorts of issues. Amit is fixing it.

PR #209, opened 2017-04-20
Kill docker option

bvandewalle:


monitor/dockermonitor/docker.go
```diff
@@ -400,13 +404,16 @@ func (d *dockerMonitor) startDockerContainer(dockerInfo *types.ContainerJSON) er
...
+			if err := d.dockerClient.ContainerStop(context.Background(), dockerInfo.ID, &timeout); err != nil {
+				log.WithFields(log.Fields{
+					"package": "monitor",
+					"error":   err.Error(),
+				}).Warn("Failed to stop bad container")
+			}
+			return fmt.Errorf("Policy cound't be set - container was killed")
 		}
-		return fmt.Errorf("Policy cound't be set - container was killed")
+		return fmt.Errorf("Policy cound't be set - container was kept alive per policy")
```
satyamsi: If we keep the container alive why are we returning error ? what will happen to this container as we move forward. I think we can get into weird states.

monitor/dockermonitor/docker.go
```diff
@@ -469,22 +476,24 @@ func (d *dockerMonitor) handleStartEvent(event *events.Message) error {
...
-		return fmt.Errorf("Cannot read container information. Killing container. ")
+			d.collector.CollectContainerEvent(&collector.ContainerRecord{
+				ContextID: contextID,
+				IPAddress: "N/A",
+				Tags:      nil,
+				Event:     collector.ContainerFailed,
+			})
+			return fmt.Errorf("Cannot read container information. Killing container. ")
+		}
+		return fmt.Errorf("Cannot read container information. Container still alive per policy. ")
```
satyamsi: Same as above why are we returning an error. If we keep container around, how would we deal with this container later ? whats the expectation ?

bvandewalle:
@satyamsi : Let me try to explain both of those.

Basically, it happens that an error occurs during the setup of the Trireme NetPolicy (Because we couldn't reach the NetNS for example as we saw recently on Kubernetes).
This flag will basically let the Container running in whatever state it was when the policy errored.

It is mainly to prevent catastrophic failure where a misconfigured agent fails to set the policy for every Container on the node, resulting on the agent killing ALL the containers on the node. If this is Kubernetes, it basically means killing the Kubernetes node completely.

So I think it makes sense depending on which level of security you want to run to be able to keep the container alive. In Kubernetes for example, the policy that we apply are only network related and it was never defined as part of the API that failure to apply them should result into killing the Container.

PR #216, opened 2017-04-24
Return errors from docker monitor start

amitlimaye:
Return error from docker monitor start so that it can be handled by caller.

### DIFF
--- a/monitor/dockermonitor/docker.go
+++ b/monitor/dockermonitor/docker.go
@@ -201,8 +201,11 @@ func (d *dockerMonitor) Start() error {
 	}).Debug("Starting the docker monitor")
 
 	// Starting the eventListener First.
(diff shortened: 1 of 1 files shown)


monitor/dockermonitor/docker.go
```diff
@@ -201,8 +201,14 @@ func (d *dockerMonitor) Start() error {
...
-	go d.eventListener()
-
+	errChan := make(chan error, 1)
+	go d.eventListener(errChan)
+	//This channel is only for catching startup errors
+	//We don't listen to this channel later or push anything to this channel
+	if eventErr := <-errChan; eventErr != nil {
+		return eventErr
+	}
+	close(errChan)
```
dstiliadis: Use defer above

PR #224, opened 2017-04-26
Counter was not increased for every flow

dstiliadis:
Counter was not increased since collection messages always give one flow.


primalmotion:
Yey!

cmd/remoteenforcer/remotecollector.go
```diff
@@ -26,6 +26,11 @@ func (c *CollectorImpl) CollectFlowEvent(record *collector.FlowRecord) {
 
 	hash := collector.StatsFlowHash(record)
 
+	// If flow event doesn't have a count make it equal to 1. At least one flow is collected
+	if record.Count == 0 {
```
primalmotion: that's where constructors are nice :)

PR #237, opened 2017-05-01
Use compact token to communicate the public key

dstiliadis:
When servers use PKI for token exchange, the public key of the server was communicated over a certificate. However, certificates tend to be big and they eat a lot of resources. In the Trireme case we don't negotiate ciphers or require all the details of the certificates. With this PR we introduce a simpler mechanism where we just communicate over the wire the EC public keys signed by the 
CA without all the overheads.


bvandewalle:
LGTM

PR #245, opened 2017-05-08
Fix: Move existing connection error message as debug

bvandewalle:


enforcer/connection/connection.go
```diff
@@ -111,7 +111,7 @@ func (c *TCPConnection) SetReported(flowState bool) {
 	state := ""
 	if repeatedReporting {
 		state = fmt.Sprintf("%t %t", c.flowLastReporting, flowState)
-		zap.L().Error("Connection reported multiple times",
+		zap.L().Debug("Connection reported multiple times",
 			zap.String("state", state))
```
dstiliadis: This is probably taken care off with the new merges. Can you try with the current master and see if the problem is still there ?

bvandewalle:
@dstiliadis  Seems it's still there :-)

PR #247, opened 2017-05-11
Using ipsets for target networks

dstiliadis:
The list of target networks can get rather big and this creates lots of overhead in the number
of ACLs that need to be managed. We can simplify this and reduce significantly the number of 
rules to effectively one by creating an ipset for the target networks. Updates are also simplified 
a lot.

This PR introduces ipsets in the target networks and solves it first for the capturing of synack packets.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ addons:
      packages:
        - libnetfilter-queue-dev
        - iptables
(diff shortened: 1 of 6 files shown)


dstiliadis:
This needs a merge. A new PR will be created.

PR #250, opened 2017-05-15
Using golang elf package directly

amitlimaye:
Using golang libs directly to find library dependencies instead of calling objdump.

### DIFF
--- a/monitor/linuxmonitor/linuxMonitor.go
+++ b/monitor/linuxmonitor/linuxMonitor.go
@@ -2,12 +2,12 @@ package linuxmonitor
 
 import (
 	"crypto/md5"
+	"debug/elf"
 	"encoding/hex"
 	"fmt"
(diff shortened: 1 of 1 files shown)


monitor/linuxmonitor/linuxMonitor.go
```diff
@@ -125,24 +125,11 @@ func findFQFN() string {
...
 	}
-
-	libraries := []string{}
-	for _, line := range strings.Split(string(out), "\n") {
-		if strings.Contains(line, "NEEDED") {
-			libraries = append(libraries, strings.TrimSpace(strings.TrimPrefix(strings.TrimSpace(line), "NEEDED")))
-		}
-	}
-
+	libraries, _ := f.ImportedLibraries()
```
dstiliadis: Please check for the error here and add error logging.

PR #254, opened 2017-05-22
Fix: Possible race condition on initial Sync

bvandewalle:


bvandewalle:
Adding @aporeto-dimitri @dstiliadis fix for potential Memory-leak

PR #255, opened 2017-05-23
Fix memory leak in nfqueue C code

dstiliadis:
This PR fixes a memory leak in the C code of the nfqueue wrapper. The allocated buffers for 
new packets were not released.


bvandewalle:
LGTM. Just tested this with Kube

PR #260, opened 2017-05-30
Handle wrap-around of version numbers

dstiliadis:
Chain versions do not need to increase forever. This creates problems for long running processes. We use XOR for the versions to limit the numbers, since we only need two.


bvandewalle:
LGTM

bvandewalle:
LGTM

PR #288, opened 2017-06-27
Duplicate Syn and SynAck Bug Fix

sibicramesh:
Consider a connection is established. If the SYN with same sequence number arrives to the enforcer, It will be dropped at the application level and will not go through the system (drop it!) provided there exists a connection for the flow. The same goes for SYNACK but, here we check for the states and then drop the packet. Also added a functionality to keep track of the number of packets dropped using atomic package.


dstiliadis:
Can we add a description with the bug we are trying to fix with this PR?

PR #304, opened 2017-07-22
Refactoring and race conditions

dstiliadis:
- Refactoring of the policy data structures to allow simpler user
- Allow multiple tags with the same key and different values 
- Solve race conditions and locks 
- Improve test coverages 
- Split files and introduce tests around policy
- Concentrate all locks to avoid messy updates


enforcer/datapath.go
```diff
@@ -20,23 +19,6 @@ import (
 	"github.com/aporeto-inc/trireme/policy"
 )
 
-// InterfaceStats for interface
```
satyamsi: I dont think we should remove stats. When we start collecting debug stats from enforcerd with the new upcoming feature, we should get these stats to see if we are dropping packets, if we are then look into debug logs. If we forward stats, the enforcer is alive and kicking.

enforcer/datapath_tcp.go
```diff
@@ -71,12 +71,10 @@ func (d *Datapath) processNetworkTCPPackets(p *packet.Packet) (err error) {
 	conn.Lock()
 	defer conn.Unlock()
 
-	d.netTCP.IncomingPackets++
```
satyamsi: I'm not in favor of removing these stats. These are per datapath stats so not that much memory and definitely not that much compute overhead.

enforcer/lookup/lookup.go
```diff
@@ -139,23 +140,35 @@ func (m *PolicyDB) AddPolicy(selector policy.TagSelector) (policyID int) {
 
 }
 
+func (m *PolicyDB) keyValueFromString(tag string) (key, value string) {
+
+	parts := strings.SplitN(tag, "=", 2)
```
satyamsi: It shouldnt be 2 but -1. Basically, I dont think we should allow = in value otherwise we can misinterpret it.

monitor/linuxmonitor/cgnetcls/cgnetcls.go
```diff
@@ -45,6 +45,11 @@ type netCls struct {
 	ReleaseAgentPath string
 }
 
+//Initialize only ince
```
satyamsi: once

enforcer/datapath.go
dstiliadis: When we actually do it, we should do it carefully. There are race conditions because of the stats, and I don't want to add locks all over the place for that.

enforcer/datapath_tcp.go
dstiliadis: They need locks. There are race conditions.

enforcer/lookup/lookup.go
dstiliadis: We actually allow that. The first equal is the one that matters. We do allocate a=b=c .. then a is the key and b=c is the value

dstiliadis:
Merged together with newer commit.

PR #308, opened 2017-07-24
Autodetect

dstiliadis:
This PR gets rid of the requirement for the dreaded Trireme Networks. It introduces an auto-detect mechanism where Trireme will autodetect based on responses whether the other side is also 
Trireme enabled or not and it will classify traffic accordingly. 

The auto-detect mechanism moves ACL processing together with auth-processing. It assumes that the other side is always Trireme enabled, unless otherwise specified. If its not, it falls back to ACL processing. This requires a new package for ACL processing that is based on hash table per prefix. The majority of ACLs are just around a couple of prefixes and this will result in every efficient lookups. 

The modifications are backwards compatible. If the Trireme Networks are defined then it will work as before. If they are not defined, then it will do auto-detect.


dstiliadis:
Merged with conntrack changes.

PR #310, opened 2017-07-25
Improve the packet tracking for Ack packets

dstiliadis:
Previous implementations relied on connmark to receive the ACK packets to user space. 

This was creating problems both with reliability as well as out-of-order delivery. With this PR
the tracking for app packets is done by us by using connmark. All packets of a flow are redirected
to user space, until authorization succeeds or we decide to drop the connection. When we are done,
we mark the flow using conntrack and the rest of the packets are forwarded directly by
the kernel. 

This actually improves performance as well since we are processing less number of packets in
user space, and we also improve security.


enforcer/datapath_tcp.go
```diff
@@ -231,6 +232,25 @@ func (d *Datapath) processApplicationTCPPacket(tcpPacket *packet.Packet, context
...
+	// We don't account for 0.0.0.0/0 external services though
+	context.Lock()
+	if action, err := context.ApplicationACLs.GetMatchingAction(tcpPacket.DestinationAddress.To4(), tcpPacket.DestinationPort); err == nil && action&policy.Accept > 0 {
+		conn.SetState(TCPData)
+		return action, nil
+	}
+	context.Unlock()
+
+	// If we have an IP in the cache for the default match, the process here.
+	if _, err := context.externalIPCache.Get(tcpPacket.DestinationAddress.String()); err == nil {
```
satyamsi: externalIPCache is accessed within context.Lock() (processNetworkSynAckPacket) and sometimes not (processApplicationSynPacket).

dstiliadis: Fixed

PR #311, opened 2017-07-25
Make trireme event processing parallel

dstiliadis:
This PR makes the event processing in Trireme parallel and allows multiple container requests to be processed at the same time, significantly improving performance. Since we still rely on exec for the iptables rules, we restrict the parallelism to the number of cores in the machine to avoid the Go runtime going wild with threads.

### DIFF
--- a/trireme.go
+++ b/trireme.go
@@ -2,6 +2,7 @@ package trireme
 
 import (
 	"fmt"
+	"runtime"
 
 	"go.uber.org/zap"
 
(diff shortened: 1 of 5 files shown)


trireme.go
```diff
@@ -413,14 +414,23 @@ func (t *trireme) Supervisor(kind constants.PUType) supervisor.Supervisor {
...
 		case req := <-t.requests:
 			zap.L().Debug("Handling Trireme Request",
 				zap.Int("type", req.reqType),
 				zap.String("contextID", req.contextID),
 			)
-			req.returnChan <- t.handleRequest(req)
+
+			sem <- true
+			go func(req *triremeRequest) {
+				req.returnChan <- t.handleRequest(req)
```
satyamsi: Is there any way two requests could be about the same context and cause out of order processing etc ?

dstiliadis: We really only care updates and creates, and an update cannot really happen before create.

PR #312, opened 2017-07-26
Fixed: Refactor ExecuteCommand

t00f:
Currently, we are passing a `map[string]interface{}` as the only argument of the `ExecuteCommand`. However, it does not provide any information about what should be included in this map. 

In order to make it clean, here is a new command named `ExecuteCommandWithParameters`. It's intended to replace the `ExecuteCommand` that I marked as **deprecated**.

If the previous command is used somewhere, please let me know. 
Otherwise, I will remove it soon.


t00f:
I have found another use of the ExecuteCommand in the **enforcerd**. [PR#196](https://github.com/aporeto-inc/enforcerd/pull/196) has been submitted and so the `ExecuteCommandWithParameters` could be renamed `ExecuteCommand`.

t00f:
Moving back to two versions of the execute command to be backward compatible with enforcers and tests.

PR #317, opened 2017-07-31
Fixed: Return correct errors

ericrpowers:
Noticed a few places were not returning the expected error, so I did a quick sweep and corrected what I noticed.

### DIFF
--- a/enforcer/acls/acls.go
+++ b/enforcer/acls/acls.go
@@ -132,7 +132,7 @@ func (c *ACLCache) AddRuleList(rules policy.IPRuleList) (err error) {
 
 	err = nil
 	for _, rule := range rules {
-		if lerr := c.AddRule(rule); err != nil {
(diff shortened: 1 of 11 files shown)


enforcer/acls/acls.go
```diff
@@ -132,7 +132,7 @@ func (c *ACLCache) AddRuleList(rules policy.IPRuleList) (err error) {
 
 	err = nil
 	for _, rule := range rules {
-		if lerr := c.AddRule(rule); err != nil {
+		if lerr := c.AddRule(rule); lerr != nil {
 			err = lerr
```
ericrpowers: As a quick question: Do we mean to continue trying to add rules even if we encounter an error?

ericrpowers: The error gets passed all the way up, so probably better to return on first error unless I am missing something? : )

PR #332, opened 2017-08-09
Default ip rules at launch will get nuked when actual policy comes th…

amitlimaye:
Added  default rule at launch of remote enforcer. This adds a rule at launch and gets removed when the actual policy comes through. The default rule install a rule to redirect packets to the first queue in the fqconfig


satyamsi (approved):

primalmotion:
what's the status on this one? If outdated, can someone close it?

amitlimaye:
Closing

PR #337, opened 2017-08-10
NEW: Support for CNI  Monitor

bvandewalle:
#### Description
CNI Monitor to receive events over RPC. This has to be used together with the CNI Client in the cni-plugin repo.

#### Test plan

> Fixes #.


bvandewalle:
@amitlimaye @dstiliadis @satyamsi 
This should be ready to be merged. 

Can I ask for a review (Probably @amitlimaye  would be the best), Would like to merge this ASAP so that we don't drift again from master

PR #352, opened 2017-08-31
New: Optional tracing mode

bvandewalle:
The user now has the option to explicitely Activate Tracing of packet header by adding the env_variable TRIREME_TRACE


bvandewalle:
#356

PR #350, opened 2017-08-29
Fixed: Do not try to collect flow event if error

ericrpowers:
#### Description
Fixes the following panic and another potential one:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x7714db]

goroutine 119 [running]:
github.com/aporeto-inc/enforcerd/collector.(*PUEventCollector).cacheFlow(0xc421187d70, 0x0)
	/tmp/build/80754af9/go/src/github.com/aporeto-inc/enforcerd/collector/statscache.go:69 +0x9b
github.com/aporeto-inc/enforcerd/collector.(*PUEventCollector).CollectFlowEvent(0xc421187d70, 0x0)
	/tmp/build/80754af9/go/src/github.com/aporeto-inc/enforcerd/collector/collector.go:60 +0x35
github.com/aporeto-inc/enforcerd/vendor/github.com/aporeto-inc/trireme/enforcer.(*nfLog).destNFLogsHandler(0xc42170ec80, 0xc4242a1c20, 0x0, 0x0)
	/tmp/build/80754af9/go/src/github.com/aporeto-inc/enforcerd/vendor/github.com/aporeto-inc/trireme/enforcer/nflog_linux.go:64 +0x97
github.com/aporeto-inc/enforcerd/vendor/github.com/aporeto-inc/trireme/enforcer.(*nfLog).(github.com/aporeto-inc/enforcerd/vendor/github.com/aporeto-inc/trireme/enforcer.destNFLogsHandler)-fm(0xc4242a1c20, 0x0, 0x0)
	/tmp/build/80754af9/go/src/github.com/aporeto-inc/enforcerd/vendor/github.com/aporeto-inc/trireme/enforcer/nflog_linux.go:38 +0x48
github.com/aporeto-inc/enforcerd/vendor/github.com/aporeto-inc/netlink-go/nflog.(*NfLog).parsePacket(0xc4218177a0, 0xc421394010, 0x87, 0xfff0, 0xac95c0, 0xc42063b250)
	/tmp/build/80754af9/go/src/github.com/aporeto-inc/enforcerd/vendor/github.com/aporeto-inc/netlink-go/nflog/nflog.go:333 +0x751
[...]

PR #355, opened 2017-09-05
Fixed: flush the cache with delay. Fixes #350. Ref #224

t00f:
#### Description
This PR adds a new method named `RemoveWithDelay` on a cache instance. When a PU is removed, the contextTracker cache is now removed with a delay of 10 sec. This is fixing the missed collection flow we had last week.

#### Test plan
Unit tests have been added on the cache object itself.

> Fixes #350

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -267,8 +267,7 @@ func (c *Cache) RemoveWithDelay(u interface{}, duration time.Duration) error {
 		return fmt.Errorf("Cannot remove item with delay - it doesn't exist")
 	}
 
-	var timer *time.Timer
-	timer = time.AfterFunc(duration, func() {
+	timer := time.AfterFunc(duration, func() {
 		if err := c.Remove(u); err != nil {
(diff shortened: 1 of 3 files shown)


bvandewalle:
What is this solving exactly?
I tried to understand the linked issue but I'm not sure I fully understand. While adding a delay to remove from the cache might work, I'm worried we are missing the underlying issue. What if 10 seconds is not "enough" in some corner cases?
@amitlimaye what is your view on this

t00f:
We have some flows that are not reported because the PU is gone and the cache has been flushed. The flow can not be collected as it could not find the PU identifier that matches. Previously, the system was raising a panic. @ericrpowers made a change to avoid the panic, but as a result, the flow is never reported.

It is clearly a timing issue: The flow is not reported yet, but the PU is already gone.

The whole idea of our discussion today with @dstiliadis is to provide a way to avoid cleaning the cache directly so the flow can be collected first.

The problem has been reported by Eric while using the soaking pipelines. Alex had some similar issue in its test case as well this morning.

t00f:
Enforcer PR has passed. Merging

Issue #359, opened 2017-09-11
Index out of range on delete namespace in Trireme


bvandewalle:
Got this error on a remote with Trireme when an ongoing delete was happening for that container:
(Triggered with Trireme-Kubernetes)

```
[0fd74fad21ab]:panic: runtime error: index out of range
[0fd74fad21ab]:
[0fd74fad21ab]:goroutine 23 [running]:
[0fd74fad21ab]:github.com/aporeto-inc/trireme-kubernetes/vendor/github.com/aporeto-inc/trireme/enforcer.(*Datapath).startApplicationInterceptor.func1(0xc4201b06e0, 0xc420531000, 0x10, 0x10, 0xc420c049ec, 0xc4216a0000)
[0fd74fad21ab]:	/home/travis/gopath/src/github.com/aporeto-inc/trireme-kubernetes/vendor/github.com/aporeto-inc/trireme/enforcer/nfq_linux.go:89 +0xaf
[0fd74fad21ab]:created by github.com/aporeto-inc/trireme-kubernetes/vendor/github.com/aporeto-inc/trireme/enforcer.(*Datapath).startApplicationInterceptor
[0fd74fad21ab]:	/home/travis/gopath/src/github.com/aporeto-inc/trireme-kubernetes/vendor/github.com/aporeto-inc/trireme/enforcer/nfq_linux.go:93 +0x424
```

bvandewalle:
Fixed. Thanks.

PR #360, opened 2017-09-11
Fix: remote out of range on Stop//Delete of a PU.

amitlimaye:
Fixes remote out of range on Stop//Delete of a PU.


bvandewalle:
Fixes #359

PR #366, opened 2017-09-19
Fix: Move the cleaning cache messages to the Trace level

bvandewalle:
Those messages are now at the trace level


bvandewalle:
Cleanup: Closing as this is not current anymore.

PR #374, opened 2017-10-06
Support for NewWithOption for Configurator

bvandewalle:
This PR uses the old logic for configurator functions but also support the NewWithOptions in the Trireme configurator.

Next step is to move all the Other  `func` to use `NewWithOptions`

### DIFF
--- a/configurator/configurator.go
+++ b/configurator/configurator.go
@@ -57,7 +57,8 @@ type TriremeOptions struct {
 	DockerSocketType string
 	DockerSocket     string
 
-	Validity time.Duration
+	Validity                time.Duration
+	ExternalIPCacheValidity time.Duration
 
 	FilterQueue *fqconfig.FilterQueue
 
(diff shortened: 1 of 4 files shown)


configurator/configurator.go
```diff
@@ -28,205 +31,301 @@ import (
...
-	secrets secrets.Secrets,
-	networks []string,
-) trireme.Trireme {
+// NewTriremeWithOptions creates all the Trireme objects based on the option struct
+func NewTriremeWithOptions(options *TriremeOptions) (*TriremeResult, error) {
 
-	if eventCollector == nil {
-		zap.L().Warn("Using a default collector for events")
-		eventCollector = &collector.DefaultCollector{}
+	enforcers := map[constants.PUType]enforcer.PolicyEnforcer{}
```
dstiliadis: Can you add some basic checks here. For example:
1. At least one of the options must be there (container or linux)
2. Local containers and process not supported together.

bvandewalle:
Thanks @dstiliadis  for the comments.
I will address them in another PR.

Please note that #361 is using the `NewWithOptions` in the background for all the existing funcs with the same signature

PR #378, opened 2017-10-12
Service context for claims 

dstiliadis:
#### Description
This PR enables services to pass context through the claims in the authorization path. The context
is transparent to Trireme, but allows the creation of services end-to-end above trireme. For example it enables the exchange of ephemeral keys for encryption or any other extensibility mechanism.

#### Test plan
Correct communication of context needs to be verified.


enforcer/datapath/tokenaccessor/tokenaccessor.go
```diff
@@ -90,20 +90,23 @@ func (t *tokenAccessor) CreateSynPacketToken(context *pucontext.PUContext, auth
...
-		T: context.Identity,
+		T:  context.Identity,
+		EK: auth.LocalServiceContext,
 	}
 
 	if context.SynToken, auth.LocalContext, err = t.getToken().CreateAndSign(false, claims); err != nil {
 		return []byte{}, nil
 	}
 
 	context.SynExpiration = time.Now().Add(time.Millisecond * 500)
```
amitlimaye: is 500 msec a little too aggressive

PR #395, opened 2017-11-16
created a separate package for uidpam

varks:
#### Description
* This PR implements a separate package for UIDPAM feature that allows one to launch protected processes based on the userid.


supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
+package uipam
```
satyamsi: whats this filename ? 

to organize the code create an interfaces.go which has an interface defined on what this module provides.

this directory should be called portset, package should be called portset and this file should be called portset.go

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+	sockStateOffset                   = 3
+	sockListeningState                = "0A"
+	hexFormat                         = 16
+	integerSize                       = 64
+	decFormat                         = 10
+)
+
+// PortSet : This provides an interface to update the
+// look up table required to program the ipset portsets.
+type PortSet interface {
```
satyamsi: move to interfaces.go

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+	AddToUIDToPortSetCache(uid string, value string) (err error)
```
satyamsi: I see a lot more functions than what we talked about. I was expecting 

type PortSet interface {
    // AddUIDPortSet is called when a user is created
    AddUID(...)
    // DeleteUIDPortSet is called when a user is deleted
    DeleteUID(...)
    // AddUIDPort is called when a new port used by a user is detected
    AddPortToUID(...)
}

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+	GetFromUIDToPortSetCache(uid string) (err error)
+	AddToUIDToPortsCache(uid string, value string) (ok bool)
+	RemoveFromUIDToPortSetCache(uid string) (err error)
+	RemoveFromUIDToPortsCache(uid string) (err error)
+	UpdatePortSet(uid string, port string)
+}
+
+// UIDCacheInstance : This type contains look up tables
```
satyamsi: Why is this public ?

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+// to help update the ipset portsets.
+type UIDCacheInstance struct {
+	UIDToPortSet *cache.Cache
+	UIDToPorts   *cache.Cache
+	UIDSet       map[string]bool
+}
+
+// NewInstance : creates a new UIDCache instance
+func NewInstance() (p *UIDCacheInstance, err error) {
```
satyamsi: Return interface not struct

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+	AddToUIDToPortsCache(uid string, value string) (ok bool)
+	RemoveFromUIDToPortSetCache(uid string) (err error)
+	RemoveFromUIDToPortsCache(uid string) (err error)
+	UpdatePortSet(uid string, port string)
+}
+
+// UIDCacheInstance : This type contains look up tables
+// to help update the ipset portsets.
+type UIDCacheInstance struct {
+	UIDToPortSet *cache.Cache
```
satyamsi: why is each of the member public ?

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+	u, err := user.Lookup(username)
+	if err != nil {
+		return "", err
+	}
+	return u.Uid, nil
+}
+
+// AddToUIDToPortsCache This adds/updates UIDToPorts cache. This cache
+// maps UID to list of active listening ports.
+func (p *UIDCacheInstance) AddToUIDToPortsCache(username string, portStr string) (ok bool, err error) {
```
satyamsi: who's intended caller for this API ?

supervisor/iptablesctrl/uidpam/update_portsets.go
satyamsi: Add descriptions on all exported functions. Not sure what are others.

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (p *UIDCacheInstance) getUserID(username string) (string, error) {
+	u, err := user.Lookup(username)
+	if err != nil {
+		return "", err
+	}
+	return u.Uid, nil
+}
+
+// AddToUIDToPortsCache This adds/updates UIDToPorts cache. This cache
```
satyamsi: is this internal or you expect to be called from outside ?

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+	RemoveFromUIDToPortSetCache(uid string) (err error)
+	RemoveFromUIDToPortsCache(uid string) (err error)
+	UpdatePortSet(uid string, port string)
+}
+
+// UIDCacheInstance : This type contains look up tables
+// to help update the ipset portsets.
+type UIDCacheInstance struct {
+	UIDToPortSet *cache.Cache
+	UIDToPorts   *cache.Cache
```
satyamsi: why are there so many caches - do we need so many. we need a memory efficient implementation.

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+	p = &UIDCacheInstance{
+		UIDToPortSet: cache.NewCache("UIDToPortSet"),
+		UIDToPorts:   cache.NewCache("UIDToPorts"),
+		UIDSet:       make(map[string]bool), // updated on portset create/delete
+	}
+
+	return p, nil
+}
+
+func (p *UIDCacheInstance) getUserID(username string) (string, error) {
```
satyamsi: who's giving us username ?

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+		ok = true
+	} else {
+		portList = append(portList, portNum)
+		p.UIDToPorts.Add(uid, portList)
+	}
+
+	return ok, nil
+}
+
+// AddToUIDToPortSetCache : This adds/updates UIDToPortSet cache. This cache
```
satyamsi: What kind of name is this - Add To UID To PortSet Cache ? Why not AddToUIDPortSetCache ? Why is this cache needed ?

supervisor/iptablesctrl/uidpam/update_portsets.go
satyamsi: Whats this name ? AddToUIDPortCache ?

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+
+// UIDCacheInstance : This type contains look up tables
+// to help update the ipset portsets.
+type UIDCacheInstance struct {
+	UIDToPortSet *cache.Cache
+	UIDToPorts   *cache.Cache
+	UIDSet       map[string]bool
+}
+
+// NewInstance : creates a new UIDCache instance
```
satyamsi: Where is the IPSet programming Interface ? Not passed in ? How do we program the ipsets ?

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+
+func startPortSetTask(p *UIDCacheInstance) {
+	t := time.NewTicker(portSetUpdateIntervalMilliseconds * time.Millisecond)
+	for range t.C {
+		// Update PortSet periodically.
+		p.updateIPPortSets()
+	}
+}
+
+func (p *UIDCacheInstance) updateIPPortSets() {
```
satyamsi: Not returning error ? If we fail there is no recovery ?

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+	scanner := bufio.NewScanner(file)
+	lineCnt := 0
+	for scanner.Scan() {
+		line := strings.Fields(scanner.Text())
+		lineCnt++
+		portList := make([]int64, 0)
+		if (lineCnt == procHeaderLineNum) || (line[sockStateOffset] != sockListeningState) {
+			continue
+		}
+		uid := line[uidFieldOffset]
```
satyamsi: potential panic if line is not long enough

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+		err := file.Close()
+		if err != nil {
+			zap.L().Warn("Failed To close /proc/net/tcp file", zap.Error(err))
+		}
+	}()
+
+	localCache := cache.NewCache("localCache")
+	scanner := bufio.NewScanner(file)
+	lineCnt := 0
+	for scanner.Scan() {
```
satyamsi: This should be like a regex match where we get a string which is port list which is then split using strings.Split as opposed to hardcoded parsing logic.

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+		puPortSetName, err := p.UIDToPortSet.Get(k) //getUIDPortSetMappings(k)
+
+		if err != nil {
+			// Not Fatal. A normal uidlogin pu will not have a port
+			continue
+		}
+
+		// ipset
+		ips := ipset.IPSet{
```
satyamsi: This should be passed in as interface to the New so this can be mocked and tested.

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+			Name: puPortSetName.(string),
+		}
+		// rearm the portset with new port List
+		for _, port := range uidPorts.([]int64) {
+			//Add an entry for 60 seconds we will rediscover ports every 60 sec
+			portNum := strconv.Itoa(int(port))
+			if adderr := ips.Add(portNum, portEntryTimeout); adderr != nil {
```
satyamsi: Whats this timeout ? We should be adding/removing.

supervisor/iptablesctrl/uidpam/update_portsets.go
```diff
@@ -0,0 +1,260 @@
...
+		return fmt.Errorf("Unable to add port to set")
+	}
+	return nil
+}
+
+// InitPortSetTask This go routine periodically scans (1s)
+// /proc/net/tcp file for listening ports and programs
+// the portsets. This worker thread is setup during datapath
+// initilisation.
+func InitPortSetTask(p *UIDCacheInstance) {
```
satyamsi: this should be done by New()

supervisor/iptablesctrl/uidpam/update_portsets.go
varks: 1) Cache to maintain UID to port mapping.( Updated by Go routine/or when user wants to add a port to UID(syn ack processing)
2) Cache to maintain UID to PortSet mapping (updated during creation/deletion of PU)
3) Need a set of active UIDs to program the ipsets periodically.

supervisor/iptablesctrl/uidpam/update_portsets.go
varks: UpdatePortSet is the interface. I will change the name to AddPortset

supervisor/iptablesctrl/uidpam/update_portsets.go
varks: The uid referenced throughout the codebase is username. We need to convert to UID in order to retrieve the portset. The /proc/net/tcp file has UID entries. Hence we need to convert username to UID.

supervisor/iptablesctrl/uidpam/update_portsets.go
varks: It could be called during syn ack processing. If the syn ack processing decides to call AddPortSet alone, then we  can make it internal.

supervisor/iptablesctrl/uidpam/update_portsets.go
varks: Names Changed.

supervisor/iptablesctrl/uidpam/update_portsets.go
varks: This is called by a go routine. Go routines cannot handle returns. If we fail, we keep trying every second. No fatal errors here.

supervisor/iptablesctrl/uidpam/update_portsets.go
varks: Removing the port from portset would require iterating over list of ports for each active uid and comparing with its previous list to determine port that needs to be removed. We know a port is no longer active when the entry no longer appears in /proc/net/tcp. 

Since we update the portset every second, we will always have a right portsets installed in the ipset.

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+	userPortSet *cache.Cache
+
+	userPortMap *cache.Cache
+}
+
+// deletes the port entry in the portset when the key uid:port
+// expires
+func expirer(c cache.DataStore, id interface{}, item interface{}) {
+
+	userPort := strings.Split(id.(string), ":")
```
satyamsi: check for length. If length < 2 we have a panic.

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+	hexFormat                         = 16
+	integerSize                       = 64
+	decFormat                         = 10
+)
+
+// portSetInstance : This type contains look up tables
+// to help update the ipset portsets.
+type portSetInstance struct {
+	userPortSet *cache.Cache
+
```
satyamsi: no need for a blank line here.

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+		portKey := uid + ":" + strconv.Itoa(int(portNum))
+		if updated := p.userPortMap.AddOrUpdate(portKey, p); updated {
+			continue
+		}
+
+		// /proc/net/tcp file contains uid. Conversion to
+		// userName is required as they are keys to lookup tables.
+		userName, err := getUserName(uid)
+		if err != nil {
+			zap.L().Warn("Error converting to username", zap.Error(err))
```
satyamsi: username is bad .. we warn and continue ? why not return ?

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+func startPortSetTask(p *portSetInstance) {
+	t := time.NewTicker(portSetUpdateIntervalMilliseconds * time.Millisecond)
+	for range t.C {
+		// Update PortSet periodically.
+		p.updateIPPortSets()
+	}
+}
+
+func (p *portSetInstance) updateIPPortSets() {
+	buffer, err := ioutil.ReadFile(procNetTCPFile)
```
satyamsi: blank line after a function creation

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		zap.L().Warn("Failed to read /proc/net/tcp file", zap.Error(err))
+		// This is a go routine, cannot return error
+		return
+	}
+
+	s := string(buffer)
+	// Ignoring the last \n
+	s = s[:len(s)-1]
```
satyamsi: not needed. check for empty line in for loop and skip. even if we need to do this use string.TrimSuffix.

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+	lines := strings.Split(s, "\n")
+
+	for cnt, line := range lines {
```
satyamsi: lines is not need - just do range strings.Split(s, "\n")

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+		line := strings.Fields(line)
+
+		if (cnt == procHeaderLineNum) || (line[sockStateOffset] != sockListeningState) {
```
satyamsi: check for length of line. If less than the max you need, continue.

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+			continue
+		}
+		uid := line[uidFieldOffset]
+		port := strings.Split(line[ipPortOffset], ":")[portOffset]
```
satyamsi: check length before accessing. i know its a specific file format but why take a chance ?

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+
+		// conver the hex port to int
+		portNum, err := strconv.ParseInt(port, hexFormat, integerSize)
+		if err != nil {
+			zap.L().Warn("Failed to convert port to Int", zap.Error(err))
+			return
```
satyamsi: why return ? continue right . otherwise one bad line makes me skip remaining lines !

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+
+		// /proc/net/tcp file contains uid. Conversion to
+		// userName is required as they are keys to lookup tables.
+		userName, err := getUserName(uid)
+		if err != nil {
+			zap.L().Warn("Error converting to username", zap.Error(err))
+		}
+
+		if err = p.addPortSet(userName, port); err != nil {
+			zap.L().Warn("Error while parsing /proc/net/tcp ", zap.Error(err))
```
satyamsi: CCP error - thats not the warning message.

cache/cache_test.go
```diff
@@ -76,10 +77,20 @@ func TestElements(t *testing.T) {
...
 
-			c.AddOrUpdate(newid, secondValue)
+			ok := c.AddOrUpdate(newid, secondValue)
 			newvalue, err := c.Get(newid)
 			So(newvalue, ShouldEqual, secondValue)
 			So(err, ShouldBeNil)
+			So(ok, ShouldBeFalse)
+		})
+
+		Convey("Given that I try to update an element in the cache, I should not get an error", func() {
```
satyamsi: fix the convey message .. .the intent of next convey message is to make sure it was an update.

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// AddPortToUser adds/updates userPortMap cache. returns
+// true if user key is already present
+func (p *portSetInstance) AddPortToUser(userName string, port string) (ok bool, err error) {
+
+	ok = false
+	key := userName + ":" + port
+
+	if _, err := p.userPortMap.Get(key); err == nil {
```
satyamsi: This block can use AddOrUpdate correct ? 

ok should be renamed to updated.

portset/portset.go
```diff
@@ -0,0 +1,225 @@
...
+
+	p.userPortSet.AddOrUpdate(userName, portset)
+	return nil
+
+}
+
+// GetUserPortSet returns the portset associated with user.
+func (p *portSetInstance) getUserPortSet(userName string) (port string, err error) {
+
+	portSetName, err := p.userPortSet.Get(userName)
```
satyamsi: 1) Merge if into same line.

2) if typecast of portSetName to string fails ? we return nil error and nil port

portset/portset.go
```diff
@@ -0,0 +1,239 @@
...
+	userPortSet *cache.Cache
+	userPortMap *cache.Cache
+}
+
+// deletes the port entry in the portset when the key uid:port
+// expires
+func expirer(c cache.DataStore, id interface{}, item interface{}) {
+
+	userPort := strings.Split(id.(string), ":")
+	portSetObject := item.(*portSetInstance)
```
satyamsi: This needs to be checked if portSetObject is not nil, if nil continue otherwise panic

portset/portset.go
```diff
@@ -0,0 +1,239 @@
...
+		userPortSet: cache.NewCache("userPortSet"),
+		userPortMap: cache.NewCacheWithExpirationNotifier("userPortMap", portEntryTimeout, expirer),
+	}
+
+	go startPortSetTask(p)
+
+	return p
+}
+
+func getUserName(uid string) (string, error) {
```
satyamsi: returning err is useless here - if string is empty, continue

PR #424, opened 2017-12-10
HandleSync for paused containers

amitlimaye:
#### Description

#### Test plan

> Fixes #.


monitor/instance/docker/monitor.go
```diff
@@ -565,8 +565,20 @@ func (d *dockerMonitor) startDockerContainer(dockerInfo *types.ContainerJSON) er
...
 	}
+	var event tevents.Event
+	if dockerInfo.State.Paused {
+		event = tevents.EventPause
+	} else if dockerInfo.State.Running {
+		event = tevents.EventStart
+	} else if dockerInfo.State.Dead {
+		event = tevents.EventStop
+	} else {
+		//We are restarting.Feeding start here. might as well be stop since we will get start notification when the
```
satyamsi: 1) Don't understand this comment more from a design process. Either we want to sent stop or start - lets make up our mind :-)

2) In case of so many if's lets use a switch.

monitor/instance/docker/monitor.go
```diff
@@ -565,8 +565,21 @@ func (d *dockerMonitor) startDockerContainer(dockerInfo *types.ContainerJSON) er
...
-
-	if err := d.config.PUHandler.HandlePUEvent(contextID, tevents.EventStart); err != nil {
+	var event tevents.Event
+	if dockerInfo.State.Paused {
+		event = tevents.EventPause
+	} else if dockerInfo.State.Running {
+		event = tevents.EventStart
+	} else if dockerInfo.State.Dead {
+		event = tevents.EventStop
+	} else {
```
satyamsi: This else is a catch all - are there other event types that could come, if yes they should be dealt with errors.

PR #427, opened 2017-12-14
Tcp nonlinux

amitlimaye:
#### Description
Don't build pipe.go for non-linux platforms
#### Test plan

> Fixes #.


bvandewalle:
partially dup #429

PR #433, opened 2017-12-19
Aws ecs

amitlimaye:
#### Description

#### Test plan

> Fixes #.

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -68,6 +68,7 @@ const (
 	DockerHostMode = "host"
 )
 
+// DockerMonitorMode defines the different modes the docker monitor can be in depending on the environment where trireme-lib is running
 type DockerMonitorMode int
 
 const (
@@ -77,6 +78,6 @@ const (
 	// KubernetesMode is a mode for docker monitor when trireme is running on host which is part of a kubernetes cluster
 	KubernetesMode
 
-	// ECSMode is a mode for docker monitor when trireme is running on host which is part of ECS/AWS cluster
-	ECSMode
+	// NoProxyMode is a mode for docker monitor when trireme is running on host which is part of ECS/AWS cluster
+	NoProxyMode
 )
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -67,3 +67,16 @@ const (
 	// DockerHostMode is the string of the network mode that indicates a host namespace
 	DockerHostMode = "host"
 )
+
+type DockerMonitorMode int
+
+const (
+	// DockerMode is a mode for docker monitor when trireme is running on host with just a docker daemon
+	DockerMode DockerMonitorMode = iota
+
+	// KubernetesMode is a mode for docker monitor when trireme is running on host which is part of a kubernetes cluster
+	KubernetesMode
+
+	// ECSMode is a mode for docker monitor when trireme is running on host which is part of ECS/AWS cluster
+	ECSMode
+)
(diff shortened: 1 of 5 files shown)


constants/constants.go
```diff
@@ -67,3 +67,16 @@ const (
...
+type DockerMonitorMode int
+
+const (
+	// DockerMode is a mode for docker monitor when trireme is running on host with just a docker daemon
+	DockerMode DockerMonitorMode = iota
+
+	// KubernetesMode is a mode for docker monitor when trireme is running on host which is part of a kubernetes cluster
+	KubernetesMode
+
+	// ECSMode is a mode for docker monitor when trireme is running on host which is part of ECS/AWS cluster
```
dstiliadis: I think that from a trireme standpoint this is essentially a NoProxyMode and not ECS mode. The ECS aspects are covered on the metadata extractors. We should change the naming convention here. We should also provide a default metadata extractor for ECS. 

In the whole PR for Trireme, nothing is ECS specific. All references to ECS should be removed and changed to NoProxyMode (or something across these lines).

internal/monitor/instance/docker/monitor.go
```diff
@@ -180,6 +193,7 @@ func DefaultConfig() *Config {
 		SocketAddress:              constants.DefaultDockerSocket,
 		SyncAtStart:                true,
 		KillContainerOnPolicyError: false,
+		ECS: false,
```
dstiliadis: See above. Nothing is ECS specific here. It is just a different mode.

internal/monitor/instance/docker/monitor.go
```diff
@@ -217,6 +230,8 @@ type dockerMonitor struct {
 	// killContainerError if enabled kills the container if a policy setting resulted in an error.
 	killContainerOnPolicyError bool
 	syncAtStart                bool
+	ECS                        bool
+	cstore                     contextstore.ContextStore
```
dstiliadis: same

amitlimaye: Renamed ECS to NoProxyMode

internal/monitor/instance/docker/monitor.go
amitlimaye: Renamed ECS to NoProxyMode

constants/constants.go
amitlimaye: Renamed ECS to NoProxyMode

internal/monitor/instance/docker/monitor.go
```diff
@@ -628,8 +690,8 @@ func (d *dockerMonitor) startDockerContainer(dockerInfo *types.ContainerJSON) er
 
 		event = tevents.EventStart
 	}
-
-	if err := d.config.PUHandler.HandlePUEvent(contextID, event); err != nil {
+	zap.L().Info("Changing Container State to:", zap.String("Event", string(event)))
```
satyamsi: remove

internal/monitor/instance/docker/monitor.go
```diff
@@ -643,12 +705,17 @@ func (d *dockerMonitor) startDockerContainer(dockerInfo *types.ContainerJSON) er
 	}
 
 	if dockerInfo.HostConfig.NetworkMode == constants.DockerHostMode {
-		if err := d.setupHostMode(contextID, runtimeInfo, dockerInfo); err != nil {
+		if err = d.setupHostMode(contextID, runtimeInfo, dockerInfo); err != nil {
 			return fmt.Errorf("unable to setup host mode for container %s: %s", contextID, err)
 		}
+		//dockerInfo.Config.Labels["$id"]
```
satyamsi: remove.

PR #435, opened 2017-12-21
New: Observe acl behavior

satyamsi:
#### Description

#### Test plan

> Fixes #.


enforcer/acls/acls.go
```diff
@@ -149,11 +149,31 @@ func (c *ACLCache) GetMatchingAction(ip []byte, port uint16) (*policy.FlowPolicy
...
+					// Check observed policies.
+					if report == nil {
+						report = p.policy
+						packet = report
+						if p.policy.ObserveAction.ObserveContinue() {
+							continue
+						} else if p.policy.ObserveAction.ObserveApply() {
+							return report, packet, nil
+						}
+					}
```
dstiliadis: I think you will miss observed policies here. They are not sorted. I think its better if we create two different tables. One for observed policies and one for actionable policies and parse in parallel, even in a go-routing.

satyamsi:
In the latest commit where I’m doing lookups in ordered form now as we actually had a bug.

We could have

x/32 deny
0/0 accept

And packets might get accepted. Now we have order from 32 down to 0. Got the functionality looking at optimizing algo/storage.

As observation will be turned on very little it has low performance overhead to have another table.

enforcer/acls/aclcache.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	report, packet, err = c.accept.getMatchingAction(ip, port, report)
+	if err == nil {
+		return
+	}
+
+	report, packet, err = c.observe.getMatchingAction(ip, port, report)
+	if err == nil {
+		return
```
dstiliadis: Isn't observe higher priority than accept ?

satyamsi: There are two observes - observes and continue (higher priority) and observe and apply (lower). The second one is to support day 0 - allow all until i get my policies sorted out whereas the observe-continue is to add incrementally on a running system.

PR #443, opened 2018-01-03
fix: update uuid.NewV4()

t00f:
Update tests to match new go.uuid version.

go.uuid defines a new prototypes for NewV1, NewV2 and NewV4 constructors.


bvandewalle:
LGTM.
The `UUID` package should be tagged (with glide or other) on whoever uses `trireme-lib` (`enforcerd`, `trireme-kubernetes`, `trireme-example`, etc etc)

PR #444, opened 2018-01-04
Support processes with multi-port definitions

dstiliadis:
#### Description
Allows the trireme lib to support processes and hosts with multi port definitions. The RPC call can
define a set of ports as 80:90. This means that ports 80 to 89 will be forwarded to this particular process. 

In the host mode this becomes very convenient. For example all ports between 1:1024 can be treated together. 

Also creates a portspec package to manage all port definitions and validation checks so that there is no code duplication and we can validate all inputs. 

The port cache will need some performance improvements, but it is isolated from the rest of the code now. It is not just a map since it needs to deal with ranges. In most cases, there is a very small number of ranges for each PU and there is no significant performance overhead. The constants in any intelligent search might be more complex than the optimization. 

#### Test plan
Unit tests have been added. 

> Fixes #.


enforcer/datapath/datapath_tcp.go
```diff
@@ -938,9 +944,9 @@ func (d *Datapath) contextFromIP(app bool, packetIP string, mark string, port st
 		return pu.(*pucontext.PUContext), nil
 	}
 
-	contextID, err := d.contextIDFromPort.Get(port)
+	contextID, err := d.contextIDFromPort.GetSpecValueFromPort(port)
```
satyamsi: Renamed GetSpecFromPort to GetSpecValueFromPort()

utils/portcache/portcache_test.go
```diff
@@ -0,0 +1,242 @@
...
+			s, err = portspec.NewPortSpec(15, 15, "range3")
+			So(err, ShouldBeNil)
+			So(p.AddUnique(s), ShouldNotBeNil)
+
+			s, err = portspec.NewPortSpec(25, 25, "range3")
+			So(err, ShouldBeNil)
+			So(p.AddUnique(s), ShouldBeNil)
+		})
+
+		Convey("When I match error'd unique entries and a valid range, I should get the valid range only", func() {
```
satyamsi: Added this testcase.

PR #448, opened 2018-01-14
Fixes for upgrade

amitlimaye:
#### Description
The PR introduces a ondataerror function on ContextStore init. 
This is called during Retrieve which will call this function in case it encounters a data store format it does not understand. 
The Remap function will convert the code to the version the current version of trireme can handle. 

The code is intended primarily used in the upgrade path allowing us to fix up the context store for each processor at startup indepently 
#### Test plan

> Fixes #.

### DIFF
--- a/internal/monitor/instance/linux/processor.go
+++ b/internal/monitor/instance/linux/processor.go
@@ -1,6 +1,7 @@
 package linuxmonitor
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"regexp"
@@ -15,12 +16,13 @@ import (
 	"github.com/aporeto-inc/trireme-lib/rpc/processor"
 	"github.com/aporeto-inc/trireme-lib/utils/cgnetcls"
 	"github.com/aporeto-inc/trireme-lib/utils/contextstore"
+	"github.com/aporeto-inc/trireme-lib/utils/portspec"
 )
 
 // StoredContext is the information stored to retrieve the context in case of restart.
 type StoredContext struct {
 	EventInfo *events.EventInfo
-	Tags      *policy.TagStore
+	Tags      *policy.TagStore `json:"Tags,omitempty"`
 }
 
 // linuxProcessor captures all the monitor processor information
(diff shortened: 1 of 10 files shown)


internal/monitor/instance/linux/processor.go
```diff
@@ -45,6 +47,30 @@ func baseName(name, separator string) string {
...
+	for index, s := range event.Services {
+		if s.Port != 0 {
+			s.Ports = &portspec.PortSpec{
+				Min: s.Port,
+				Max: s.Port,
+			}
+		}
+		event.Services[index].Ports = s.Ports
+	}
+	if fixedData.(*StoredContext).Tags == nil {
```
satyamsi: 1) Potential panic in conversion
2) fixedData.(*StoredContext) should be converted once and temp var should be used after that.

internal/monitor/instance/linux/processor.go
```diff
@@ -293,7 +319,7 @@ func (l *linuxProcessor) ReSync(e *events.EventInfo) error {
 		reacquired = append(reacquired, eventInfo.PUID)
 
 		// Synchronize
-		if l.config.SyncHandler != nil {
+		if l.config.SyncHandler != nil && storedContext.Tags != nil && len(storedContext.Tags.GetSlice()) != 0 {
```
satyamsi: Will storedContext.Tags ever be nil given that we have remap ? Maybe we should have a check that once we retrieve the tags are not nil and not have the rest of the code keep checking this.

internal/monitor/instance/uid/processor.go
```diff
@@ -61,6 +63,29 @@ func baseName(name, separator string) string {
...
+	for index, s := range event.Services {
+		if s.Port != 0 {
+			s.Ports = &portspec.PortSpec{
+				Min: s.Port,
+				Max: s.Port,
+			}
+		}
+		event.Services[index].Ports = s.Ports
+	}
+	if fixedData.(*StoredContext).Tags == nil {
```
satyamsi: Same as above.

utils/contextstore/contextstore.go
```diff
@@ -69,11 +71,20 @@ func (s *store) Retrieve(contextID string, context interface{}) error {
...
+
 	if err != nil {
 		return fmt.Errorf("unable to retrieve context from store: %s", err)
 	}
 
 	if err = json.Unmarshal(data, context); err != nil {
+		if s.dataErrorHandler != nil {
+			if err = s.dataErrorHandler(string(data), context); err == nil {
+				if err = s.Store(contextID, context); err != nil {
+					return fmt.Errorf("Remap attempt failed with error %s", err)
```
satyamsi: In case of error, should we remove or not. Today we remove because now we can keep running this but we will never recover from this error no matter how many restarts we do.

utils/contextstore/contextstore_test.go
```diff
@@ -124,7 +124,7 @@ func TestRemove(t *testing.T) {
 
 func TestWalk(t *testing.T) {
 
-	cstore := NewFileContextStore(storebasePath)
+	cstore := NewFileContextStore(storebasePath, nil)
```
satyamsi: Not a single test to test positive/negative remap.

PR #459, opened 2018-01-19
resync in case of docker daemon restart 

amitlimaye:
#### Description
Resync containers if the docker daemon restarts.
#### Test plan

> Fixes #.


dstiliadis:
What was this solving and is it still needed? Either close or update to the current state.

PR #485, opened 2018-03-09
Fix: Error propagation in the remote and Correct PSKType for PSK secrets

bvandewalle:
This PR corrects:

1) Correct error propagation for the remote enforcer (took me 2 hours of debug to see that error)
2) the PSK Secret type to correctly identify itself

### DIFF
--- a/controller/internal/enforcer/applicationproxy/applicationproxy.go
+++ b/controller/internal/enforcer/applicationproxy/applicationproxy.go
@@ -66,7 +66,8 @@ func NewAppProxy(tp tokenaccessor.TokenAccessor, c collector.EventCollector, puF
 		return nil, err
 	}
 
-	if ok := systemPool.AppendCertsFromPEM(s.PublicSecrets().CertAuthority()); !ok {
+	// We append the CA only if we are not in PSK mode as it doesn't provide a CA.
+	if ok := systemPool.AppendCertsFromPEM(s.PublicSecrets().CertAuthority()); !ok && (s.PublicSecrets().SecretsType() != secrets.PSKType) {
 		return nil, fmt.Errorf("Cannot append ca chain")
 	}
 
--- a/controller/internal/enforcer/applicationproxy/applicationproxy.go
+++ b/controller/internal/enforcer/applicationproxy/applicationproxy.go
@@ -67,8 +67,10 @@ func NewAppProxy(tp tokenaccessor.TokenAccessor, c collector.EventCollector, puF
 	}
 
 	// We append the CA only if we are not in PSK mode as it doesn't provide a CA.
-	if ok := systemPool.AppendCertsFromPEM(s.PublicSecrets().CertAuthority()); !ok && (s.PublicSecrets().SecretsType() != secrets.PSKType) {
-		return nil, fmt.Errorf("Cannot append ca chain")
+	if s.PublicSecrets().SecretsType() != secrets.PSKType {
+		if ok := systemPool.AppendCertsFromPEM(s.PublicSecrets().CertAuthority()); !ok {
+			return nil, fmt.Errorf("error while adding provided CA")
+		}
 	}
 
 	return &AppProxy{
--- a/controller/internal/enforcer/proxy/enforcerproxy.go
+++ b/controller/internal/enforcer/proxy/enforcerproxy.go
@@ -69,7 +69,7 @@ func (s *ProxyInfo) InitRemoteEnforcer(contextID string) error {
 	}
 
 	if resp.Status != "" {
-		return fmt.Errorf("failed to initialize remote enforcer: status: %s", resp.Status)
+		zap.L().Warn("received status while initializing the remote enforcer", zap.String("contextID", resp.Status))
 	}
 
 	s.Lock()
--- a/controller/internal/enforcer/proxy/enforcerproxy.go
+++ b/controller/internal/enforcer/proxy/enforcerproxy.go
@@ -68,6 +68,10 @@ func (s *ProxyInfo) InitRemoteEnforcer(contextID string) error {
 		return fmt.Errorf("failed to initialize remote enforcer: status: %s: %s", resp.Status, err)
 	}
 
+	if resp.Status != "" {
+		return fmt.Errorf("failed to initialize remote enforcer: status: %s", resp.Status)
+	}
+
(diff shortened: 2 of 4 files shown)


bvandewalle:
@dstiliadis can you review//approve please ? Thanks

controller/internal/enforcer/proxy/enforcerproxy.go
```diff
@@ -68,6 +68,10 @@ func (s *ProxyInfo) InitRemoteEnforcer(contextID string) error {
 		return fmt.Errorf("failed to initialize remote enforcer: status: %s: %s", resp.Status, err)
 	}
 
+	if resp.Status != "" {
```
bvandewalle, line 71: We were only checkin that the RPC return was errored and only then the status. 
Now if the status is not empty, we display a warning, but I think this should always result in an error

controller/pkg/remoteenforcer/remoteenforcer_linux.go
```diff
@@ -114,7 +115,7 @@ func (s *RemoteEnforcer) setupEnforcer(req rpcwrapper.Request) (err error) {
 		payload.ExternalIPCacheTimeout,
 		payload.PacketLogs,
 	); err != nil || s.enforcer == nil {
-		return errors.New("unable to setup enforcer: we don't know as this function does not return an error")
+		return fmt.Errorf("Error while initializing remote enforcer, %s", err)
```
bvandewalle, line 118: The error here is actually meaningful and can be consumed, so we should return it

controller/pkg/remoteenforcer/remoteenforcer_linux.go
```diff
@@ -172,7 +173,7 @@ func (s *RemoteEnforcer) InitEnforcer(req rpcwrapper.Request, resp *rpcwrapper.R
 
 	if err := s.setupEnforcer(req); err != nil {
 		resp.Status = err.Error()
-		return nil
+		return fmt.Errorf(resp.Status)
```
bvandewalle, line 182: Why not returning it as an error ? Previously the status would be completely ignored

bvandewalle (commented):
Trying to explain some of my changes inline.

controller/internal/enforcer/applicationproxy/applicationproxy.go
```diff
@@ -66,7 +66,8 @@ func NewAppProxy(tp tokenaccessor.TokenAccessor, c collector.EventCollector, puF
 		return nil, err
 	}
 
-	if ok := systemPool.AppendCertsFromPEM(s.PublicSecrets().CertAuthority()); !ok {
+	// We append the CA only if we are not in PSK mode as it doesn't provide a CA.
+	if ok := systemPool.AppendCertsFromPEM(s.PublicSecrets().CertAuthority()); !ok && (s.PublicSecrets().SecretsType() != secrets.PSKType) {
 		return nil, fmt.Errorf("Cannot append ca chain")
 	}
```
dstiliadis: Don't return an error here. If there is no CA, we just use the system pool. Just check if the secrets type is not PSK and ignore it if it is.

controller/internal/enforcer/proxy/enforcerproxy.go
```diff
@@ -68,6 +68,10 @@ func (s *ProxyInfo) InitRemoteEnforcer(contextID string) error {
...
+		zap.L().Warn("received status while initializing the remote enforcer", zap.String("contextID", resp.Status))
+	}
+
 	s.Lock()
```
dstiliadis, line 75: What is the scenario where the status is interesting if there is no error? I actually think that we should only check the status if there is an error and insert it in the error message so that callers will know why it failed.

controller/pkg/remoteenforcer/remoteenforcer_linux.go
dstiliadis: Yes, return an error here.

dstiliadis (changes requested):

controller/internal/enforcer/applicationproxy/applicationproxy.go
bvandewalle: This should be what this code is doing. I will  change it to make it more clear.

controller/internal/enforcer/proxy/enforcerproxy.go
bvandewalle: While reading the RPC code, I got confused because we return an error but also a status, which is always redundant, and usually the same. Why are those separate, ans why do we need a status?

From what I can see the only time we use a status is also when we return an error.

controller/pkg/remoteenforcer/remoteenforcer_linux.go
amitlimaye: stop evaluating resp.Status on the client side if you return an error here

controller/internal/enforcer/proxy/enforcerproxy.go
amitlimaye: The status will not be populated if there is an error. The response object is not sent back to the client if an error is returned from this method

bvandewalle: that's not what I'm seeing. Both the error and status are successfully returned. Hence my question of why do we need those ?

bvandewalle:
So, we have only two spots where we do not return an error but set a status only:
```
	if err != nil {
		zap.L().Error("Remote enforcer failed: unable to identify namespace",
			zap.String("nsErr", nsEnterState),
			zap.String("nsLogs", nsEnterLogMsg),
			zap.Error(err),
		)
		resp.Status = fmt.Sprintf("Remote enforcer failed: unable to identify namespace: %s", err)

		// Dont return error to close RPC channel
	}
```

Why is that ?

controller/internal/enforcer/proxy/enforcerproxy.go
amitlimaye: net/rpc doc says if you return non-nil error response object will not be populated. 
The idea was to always return nil from remote enforcer and only use the resp.Status to return errors from remote enforcer and use the err to indicate rpc channel failure or encoder issues

amitlimaye (approved):

dstiliadis (approved):

PR #495, opened 2018-03-16
Kubernetes monitor

bvandewalle:
This PR introduces a Monitor for Kubernetes that will only send POD events upstream.

It also connects to Kubernetes API in order to complete and merge the tags.

It reuses the Docker Monitor as intermediate and filter its events.

### DIFF
--- a/monitor/internal/kubernetes/cache.go
+++ b/monitor/internal/kubernetes/cache.go
@@ -13,12 +13,17 @@ type podCacheEntry struct {
 	runtime policy.RuntimeReader
 	// The latest known reference to the pod received from Kubernetes API
 	pod *api.Pod
+
+	// Lock for the specific entry
+	sync.RWMutex
 }
 
 // Cache keeps all the state needed for the integration.
 type cache struct {
 	// contextIDCache keeps a mapping between a POD/Namespace name and the corresponding contextID from Trireme.
 	podCache map[string]*podCacheEntry
+
+	// Lock for the whole cache
 	sync.RWMutex
 }
 
--- a/monitor/internal/kubernetes/cache.go
+++ b/monitor/internal/kubernetes/cache.go
@@ -32,3 +32,10 @@ func newCache() *cache {
 func kubePodIdentifier(podName string, podNamespace string) string {
 	return podNamespace + "/" + podName
 }
+
+func (c *cache) getPodFromCache(podNamespace string, podName string) *podCacheEntry {
+	c.Lock()
+	defer c.Unlock()
+	kubeIdentifier := kubePodIdentifier(podName, podNamespace)
+	return c.podCache[kubeIdentifier]
+}
--- a/monitor/internal/kubernetes/cache.go
+++ b/monitor/internal/kubernetes/cache.go
@@ -34,7 +34,8 @@ type cache struct {
 // NewCache initialize a cache
 func newCache() *cache {
 	return &cache{
-		podCache: map[string]*podCacheEntry{},
+		podCache:  map[string]*podCacheEntry{},
+		puidCache: map[string]string{},
 	}
 }
 
@@ -42,8 +43,8 @@ func kubePodIdentifier(podName string, podNamespace string) string {
 	return podNamespace + "/" + podName
 }
 
-// getOrCreatePodFromCache locks the cache in order to return the pod cache entry if found, or create it if not found
-func (c *cache) getOrCreatePodByKube(podNamespace string, podName string) *podCacheEntry {
+// createPodEntry locks the cache in order to return the pod cache entry if found, or create it if not found
+func (c *cache) createPodEntry(podNamespace string, podName string, puID string, runtime policy.RuntimeReader) *podCacheEntry {
 	c.Lock()
 	defer c.Unlock()
 
@@ -53,7 +54,31 @@ func (c *cache) getOrCreatePodByKube(podNamespace string, podName string) *podCa
 		cacheEntry = &podCacheEntry{}
 		c.podCache[kubeIdentifier] = cacheEntry
 	}
+	cacheEntry.puID = puID
+	cacheEntry.runtime = runtime
+
+	c.puidCache[puID] = kubeIdentifier
+
 	return cacheEntry
+}
+
+func (c *cache) updatePodEntry(podNamespace string, podName string, pod *api.Pod) (*podCacheEntry, error) {
+	c.Lock()
+	defer c.Unlock()
+
+	kubeIdentifier := kubePodIdentifier(podName, podNamespace)
+	cacheEntry, ok := c.podCache[kubeIdentifier]
+	if !ok {
+		cacheEntry = &podCacheEntry{}
+		c.podCache[kubeIdentifier] = cacheEntry
+	}
+
+	cacheEntry.Lock()
+	defer cacheEntry.Unlock()
+
+	cacheEntry.pod = pod
+
+	return cacheEntry, nil
 }
 
 // getOrCreatePodFromCache locks the cache in order to return the pod cache entry if found, or create it if not found
--- a/monitor/internal/kubernetes/cache.go
+++ b/monitor/internal/kubernetes/cache.go
@@ -12,17 +12,11 @@ type puidCacheEntry struct {
 
 	// The latest reference to the runtime as received from DockerMonitor
 	runtime policy.RuntimeReader
-
-	// Lock for the specific entry
-	sync.RWMutex
 }
 
 type podCacheEntry struct {
 	// puIDs us a map containing a link to all the containers currently known to be part of that pod.
 	puIDs map[string]bool
-
-	// Lock for the specific entry
-	sync.RWMutex
 }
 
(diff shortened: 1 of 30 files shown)


monitor/internal/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	// If IP is empty, wait for an UpdatePodEvent with the Actual PodIP. Not ready to be activated now.
+	if pod.Status.PodIP == "" {
+		return nil, nil
+	}
+	// If Pod is running in the hostNS, no activation (not supported).
+	if pod.Status.PodIP == pod.Status.HostIP {
+		return nil, nil
+	}
```
dstiliadis: we can support that.

dstiliadis:
Could we do that as just a metadataextractor rather than a new monitor? The main thing that it does is retrieve the metadata from the Kubernetes. Docker monitor gets the events and calls a metadataextractor to create the runtime that collects the metadata both from the docker event as well as Kubernetes. The extractor can also register watchers for updates. In this way we don't need to repeat the logic of the monitor. Mostly concerned about resyncs.

bvandewalle:
I did it as an extractor initially, but after trying it I came to the conclusion that it is better to use it as a Monitor.
Here are the pros and cons:

DockerMonitor with Kubernetes metadata Extractor (similar to what we have today almost):
- pro: Not a lot of changes
- big CON: Every container event not relevant to Kubernetes is still sent up to the handler
- big CON: Does not support explicit update of a PU from the metadataExtracotor. This is needed if a PU updates its tags.

KubernetesMonitor
- Pro: the PU event that is sent up is tailored to a Pod from Kubernetes, and not to the Docker container
- Pro: Will be more easily adapted to the eventual CRI-O (or other) container engine
- Con: Relies on another monitor

The resync mechanism is very simple, it is simply calling the DockerMonitor resync which resends all the events to the Kubernetes monitor.
The Kubernetes Monitor is simply a passthrough monitor using the DockerMonitor and selecting which events should be sent upstream, also filtering and adding the tags.

monitor/internal/kubernetes/kubernetes.go
bvandewalle: indeed, will add it as an option.

monitor/internal/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,135 @@
...
+
+	pod, err := m.kubernetesClient.Pod(podName, podNamespace)
+	if err != nil {
+		return nil, fmt.Errorf("Couldn't get labels for pod %s : %v", podName, err)
+	}
+
+	// If IP is empty, wait for an UpdatePodEvent with the Actual PodIP. Not ready to be activated now.
+	if pod.Status.PodIP == "" {
+		return nil, nil
+	}
```
dstiliadis: I was thinking about it. We don't really care if IP is empty. We can use 0.0.0.0/0 .. Its not used anywhere, other than stats.

bvandewalle: yes I need to change it , it's a remnant from the old Trireme-Kubernetes.

We indeed do not care about the IP, but it is an indicator that the pod is not "ready" yet.

What we need to do now is to cache the Pod//PuRuntime until it is in a ready state to be sent up.

monitor/internal/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,90 @@
...
+	zap.L().Debug("pod added event", zap.String("name", addedPod.GetName()), zap.String("namespace", addedPod.GetNamespace()))
+
+	// This event is not needed as the trigger is the  DockerMonitor event
+	// The pod obejct is cached in order to reuse it and avoid an API request possibly laster on
+
+	// _, err := m.cache.updatePodEntry(addedPod.GetNamespace(), addedPod.GetName(), addedPod)
+	// if err != nil {
+	//	return fmt.Errorf("error updating cache entry %s", err)
+	//}
+
```
dstiliadis: Let's remove commented out code.

monitor/internal/kubernetes/monitor.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// As the Kubernetes monitor depends on the DockerMonitor, we setup the Docker monitor first
+	dockerMon := dockermonitor.New()
+	dockerMon.SetupHandlers(processorConfig)
+
+	dockerConfig := dockermonitor.DefaultConfig()
+	dockerConfig.EventMetadataExtractor = kubernetesconfig.DockerExtractor
+
+	// we use the defaultconfig for now
+	if err := dockerMon.SetupConfig(nil, dockerConfig); err != nil {
```
dstiliadis, line 77: This will be an issue with the bigger integrations and the APIs. Isn't this propagated through the instantiation of hte library (I mean the defaultConfig for docker)?

monitor/internal/kubernetes/monitor.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// Run starts the monitor.
+func (m *KubernetesMonitor) Run(ctx context.Context) error {
+	if m.kubeClient == nil {
+		return fmt.Errorf("kubernetes client is not initialized correctly")
+	}
+
+	go m.podController.Run(m.podControllerStop)
+	initialPodSync := make(chan struct{})
```
dstiliadis: If context is not massed to the podController.Run how is it stopping ?

monitor/internal/kubernetes/monitor.go
```diff
@@ -0,0 +1,142 @@
...
+// Resync requests to the monitor to do a resync.
+func (m *KubernetesMonitor) Resync(ctx context.Context) error {
+	// TODO: Redifine this interface ?
+	return nil
+}
+
+// ReSync ???
+func (m *KubernetesMonitor) ReSync(ctx context.Context) error {
+	return m.dockerMonitor.ReSync(ctx)
+}
```
dstiliadis: Why is there Resync and ReSync ?

policy/interfaces.go
```diff
@@ -24,6 +24,9 @@ type RuntimeReader interface {
 	// Name returns the process name of the Runtime.
 	Name() string
 
+	// NSPath returns the process name of the Runtime.
+	NSPath() string
+
```
dstiliadis: That doesn't seem to match the comment. NSPath returns the process name of the Runtime ? Where is this used?

monitor/internal/kubernetes/monitor.go
```diff
@@ -0,0 +1,142 @@
...
+	m.localNode = kubernetesconfig.Nodename
+	kubeClient, err := NewKubeClient(kubernetesconfig.Kubeconfig)
+	if err != nil {
+		return fmt.Errorf("kubernetes client instantiation error: %s", err.Error())
+	}
+	m.kubeClient = kubeClient
+
+	m.enableHostPods = kubernetesconfig.EnableHostPods
+	m.kubernetesExtractor = kubernetesconfig.KubernetesExtractor
+
```
dstiliadis, line 93: Have we tested with host pods ?

monitor/internal/kubernetes/cache.go
```diff
@@ -0,0 +1,147 @@
...
+	// popuidCache keeps a mapping between a PUID and the corresponding puidCacheEntry.
+	puidCache map[string]*puidCacheEntry
+
+	// podCache keeps a mapping between a POD/Namespace name and the corresponding podCacheEntry.
+	podCache map[string]*podCacheEntry
+
+	// Lock for the whole cache
+	sync.RWMutex
+}
+
```
dstiliadis, line 41: It is declared as a RW mutex, but below it is only used a plain mutex. No RLock() RUnlock(). I also think that if we used the sync Map package of go it would be better. Trying to migrate all caches to this. In any case, let's use RLock() in the read operations below.

dstiliadis (changes requested):

monitor/internal/kubernetes/monitor.go
bvandewalle: https://github.com/aporeto-inc/trireme-lib/blob/master/monitor/interfaces.go

I'm maybe missing something but Monitor seems to be required to implement both of those interface. And for some reason we got a Resync and a ReSync.

policy/interfaces.go
bvandewalle: Will update the comments.

Two reasons:
1) it is used here
https://github.com/aporeto-inc/enforcerd/pull/545/files#diff-f46d3978f0ef10ba2d41f0994134fd57R419

2) All the parameters can be read from that interface. Why not NSPath ? Seems like an ommition

monitor/internal/kubernetes/monitor.go
bvandewalle: Good point, this should be handled by Context.

monitor/internal/kubernetes/cache.go
bvandewalle: Done, Thanks, I didn't know about `Rlock` and `RUnlock`

dstiliadis (approved):

PR #500, opened 2018-03-28
Allow logging to file and console simultaneously

satyamsi:
#### Description

#### Test plan

> Fixes #.


controller/helpers.go
```diff
@@ -39,10 +39,9 @@ func GetLogParameters() (logToConsole bool, logID string, logLevel string, logFo
 
 	if console := os.Getenv(constants.EnvLogToConsole); console == constants.EnvLogToConsoleEnable {
```
primalmotion: why are we doing getenv here? this is uncontrollable. This should come from the configuration

amitlimaye (approved):

PR #505, opened 2018-04-12
New: Generate a runtime for All events. Including Pause, Unpause, Stop, Destroy

bvandewalle:
This PR adds uniformity to all the events types for the Docker Monitor.

Upto now, only the Start and Create events were generating metadata.
For all the other events, we were sending a PURuntime with Defaults. That doesn't really make sense and disallow any type of Processing to the upper level Resolvers if they don't keep a local cache.

This PR adds processing of the Runtime for all the event types.

### DIFF
--- a/monitor/internal/docker/monitor.go
+++ b/monitor/internal/docker/monitor.go
@@ -376,34 +376,44 @@ func (d *DockerMonitor) extractMetadata(dockerInfo *types.ContainerJSON) (*polic
 	return extractors.DefaultMetadataExtractor(dockerInfo)
 }
 
-// handleCreateEvent generates a create event type. We extract the metadata
-// and start the policy resolution at the create event. No need to wait
-// for the start event.
-func (d *DockerMonitor) handleCreateEvent(ctx context.Context, event *events.Message) error {
-
-	puID, err := puIDFromDockerID(event.ID)
-	if err != nil {
-		return err
-	}
+// containerAndRuntimeFromEvent is an helper function that returns the Runtime (from the metadata extractor) and Container(from Docker)
+func (d *DockerMonitor) containerAndRuntimeFromEvent(ctx context.Context, event *events.Message) (*types.ContainerJSON, *policy.PURuntime, error) {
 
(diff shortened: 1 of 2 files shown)


monitor/internal/docker/monitor.go
```diff
@@ -508,14 +526,21 @@ func (d *DockerMonitor) handleDestroyEvent(ctx context.Context, event *events.Me
...
 		return err
 	}
 
-	return d.config.Policy.HandlePUEvent(ctx, puID, tevents.EventPause, policy.NewPURuntimeWithDefaults())
+	// The runtime is inferred from the cache if possible.
+	// If impossible, the runtime is set to a default empty Runtime.
+	runtime := d.runtimeCache.getRuntime(puID)
+	if runtime == nil {
+		runtime = policy.NewPURuntimeWithDefaults()
+	}
```
Dogild, line 541: Why do we create a new default runtime here ? Would it be more logic to create a new runtime by calling the method `containerAndRuntimeFromEvent` ?

bvandewalle: 3 reasons:

1) The PU Runtime didn't change so no need to regenerate it
2) It is in the cache so way more efficient to get it from there
3) In some event cases (such as stop or destroy), the PU might be gone from Docker already, in which case we can only get the runtime from cache anymore

bvandewalle: The runtime is only set to default if it cannot get it from  the cache, which means we never saw the PU before and most likely docker doesnt know about it

dstiliadis:
I am not sure this is needed. Main concern is that we are keeping state in the monitor again and we went through hoops to avoid that. Second issue is that there is no need for metadata in stop/pause. Third reason is that we external extractors this introduces again latency.

dstiliadis (changes requested):
See comments below. I am not sure we should re-introduce state in the monitor. What is the reason that a resolver should care about metadata on a stop/pause? If they care, they should maintain state themselves.

PR #510, opened 2018-04-18
Validate interface type for certificates

dstiliadis:
#### Description
Solves a crash when the signing method of the JWT uses a different certificate type than the one provided for the service.


t00f (approved):
Looks good. It fixes the panic I have seen today. Thanks!

PR #518, opened 2018-04-25
Backport cert renewal fix to 2.0.0

varks:


controller/internal/enforcer/nfqdatapath/datapath.go
```diff
@@ -322,6 +322,8 @@ func (d *Datapath) Run(ctx context.Context) error {
 
 // UpdateSecrets updates the secrets used for signing communication between trireme instances
 func (d *Datapath) UpdateSecrets(token secrets.Secrets) error {
+
+	d.secrets = token
```
satyamsi, line 326: Is token a pointer or a copy. Given that we have so many concurrency issues, I want to make sure storing a ref doesnt cause issues.

satyamsi (approved):

PR #525, opened 2018-05-11
Fix and update flow stats for TCP and HTTP connections

dstiliadis:
#### Description
Fixes the flow statistics reporting for TCP flows and provides the service and policy parameters. Cleans up the flow statistics reporting for HTTP requests.

### DIFF
--- a/controller/internal/enforcer/applicationproxy/applicationproxy.go
+++ b/controller/internal/enforcer/applicationproxy/applicationproxy.go
@@ -107,7 +107,7 @@ func (p *AppProxy) Enforce(ctx context.Context, puID string, puInfo *policy.PUIn
 	defer p.Unlock()
 
 	// First update the caches with the new policy information.
-	apicache, dependentCache, jwtcache, caPool := buildCaches(puInfo.Policy.ExposedServices(), puInfo.Policy.DependentServices())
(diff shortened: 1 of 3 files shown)


controller/internal/enforcer/applicationproxy/applicationproxy.go
```diff
@@ -382,13 +383,20 @@ func serviceTypeToApplicationListenerType(serviceType policy.ServiceType) protom
 	}
 }
 
-func buildCaches(exposedServices, dependentServices policy.ApplicationServicesList) (map[string]*urisearch.APICache, map[string]*urisearch.APICache, map[string]*x509.Certificate, [][]byte) {
+func buildCaches(exposedServices, dependentServices policy.ApplicationServicesList) (map[string]*urisearch.APICache, map[string]*urisearch.APICache, map[string]*x509.Certificate, [][]byte, map[int]string) {
 	apicache := map[string]*urisearch.APICache{}
```
primalmotion, line 387: Shouldn’t you use an lru cache? Those maps may grow forever

dstiliadis: No. they get replaced with every policy update.

PR #564, opened 2018-08-01
Use ACL cache to resolve udp external services

varks:


amitlimaye (approved):

controller/internal/enforcer/nfqdatapath/datapath_udp.go
```diff
@@ -638,3 +659,48 @@ func (d *Datapath) processNetworkUDPAckPacket(udpPacket *packet.Packet, context
...
+	_, action, err = context.NetworkUDPACLPolicyFromAddr(p.DestinationAddress.To4(), p.SourcePort)
+	if err == nil {
+		return action, err
+	}
+
+	report, action, err := context.ApplicationUDPACLPolicyFromAddr(p.DestinationAddress.To4(), p.DestinationPort)
+	if err != nil {
+		return nil, err
+	}
+
```
dstiliadis, line 680: Can't we just check the protocol and call one of the above, rather than both?

controller/internal/enforcer/nfqdatapath/datapath_udp.go
```diff
@@ -638,3 +659,48 @@ func (d *Datapath) processNetworkUDPAckPacket(udpPacket *packet.Packet, context
...
+	if err == nil {
+		return action, nil
+	}
+
+	// check for network ACLS
+	report, action, err := context.NetworkUDPACLPolicyFromAddr(p.SourceAddress.To4(), p.DestinationPort)
+	if err != nil {
+		return nil, err
+	}
+
```
dstiliadis, line 703: Don't we know the direction? I can see this creating problems if I use 0.0.0.0. Why check for both directions?

dstiliadis (changes requested):
See above for some suggestions.

dstiliadis (approved):

PR #608, opened 2018-09-04
Use iptables restore for ACL programming

dstiliadis:
#### Description
This PR changes the programming of ACLs to use iptables restore. This reduces the number of calls to the kernel and the number of exec calls and significantly improves performance.

It is implemented as new iptables provider that caches all the commands provided by the existing code and commits the output together through an explicit commit message.


controller/pkg/aclprovider/iptablesprovider.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+// Append will append the provided rule to the local cache or call
+// directly the iptables command depending on the table.
+func (b *BatchProvider) Append(table, chain string, rulespec ...string) error {
+	b.Lock()
+	defer b.Unlock()
+
+	if _, ok := b.batchTables[table]; !ok {
+		return b.ipt.Append(table, chain, rulespec...)
```
varks, line 84: Is this because for tables other than mangle, (nat etc) we use old method of programming iptables?

controller/pkg/aclprovider/iptablesprovider.go
```diff
@@ -0,0 +1,299 @@
...
+	if _, ok := b.rules[table][chain]; !ok {
+		b.rules[table][chain] = []string{}
+	}
+
+	rule := strings.Join(rulespec, " ")
+
+	if pos == 1 {
+		b.rules[table][chain] = append([]string{rule}, b.rules[table][chain]...)
+	} else if pos > len(b.rules[table][chain]) {
+		b.rules[table][chain] = append(b.rules[table][chain], rule)
```
varks, line 126: Clarification: Converting all insert rules to append since iptables-restore buffer only supports append. Is that the case?

controller/pkg/aclprovider/iptablesprovider.go
dstiliadis: yes. it's selective

controller/pkg/aclprovider/iptablesprovider.go
dstiliadis: correct. The buffer has the explicit sequence that you want and sends only appends.

PR #617, opened 2018-09-17
Capture pcap

amitlimaye:
#### Description
a way to capture packets for a particular flow on the enforcer and generate a pcap file which can be used in wireshark
#### Test plan

> Fixes #.

### DIFF
--- a/controller/pkg/packet/capture.go
+++ b/controller/pkg/packet/capture.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"os"
-	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


controller/pkg/packet/capture.go
```diff
@@ -0,0 +1,149 @@
...
+	threadNumToFlow = make([]*captureFlow, numQueues)
+	if err != nil {
+		return
+	}
+
+	go updateFlows()
+}
+
+func updateFlows() {
+	var lastModified time.Time
```
amitlimaye, line 47: this can be done with a ReadFile and os.Stat you don't need to track handles that way

PR #643, opened 2018-10-11
Do not delete cgroup for network host services.

varks:
cgroup mark is not programmed for network host services and hence shouldn't be removed. On removal of network host services, cgroup shouldn't be written "0" mark for network only host services. It should be written to 0 and removed only when a host pu (both incoming/outgoing in old parlance) is removed.


varks:
```
[
  {
    "project": "",
    "component": "trireme-lib",
    "pr-id": "643",
    "commit-sha": "7ec85ae72b5a34d08489b185b9698420376ca121"
  }
]
```

amitlimaye (approved):

PR #652, opened 2018-10-17
check if both ends are loopback and don't update connmark on applicat…

0xmchadha:
…ion ack. Let network ack be processed to update connamrk (#648)

Check if both ends are loopback ip's and don't update connmark on application ack

#### Description

#### Test plan

> Fixes #.


dstiliadis:
Closing this. It is for 3.3.0

PR #673, opened 2018-10-28
Fix crash when connection doesn't have a remote address

dstiliadis:
Fixes a rare crash when net.Conn doesn't have the remote address.

https://github.com/golang/go/issues/23022


primalmotion (changes requested):
You should save a call by putting RemoteAddr() result in a variable

dstiliadis:
@primalmotion fixed.

primalmotion (approved):

PR #695, opened 2018-11-28
Retry name lookup on failure after 500 ms

0xmchadha:
While doing DNS lookups on behalf of the PU, if the failure occurs, retry after 500ms.


varks (approved):

controller/pkg/pucontext/pucontext.go
```diff
@@ -116,44 +116,61 @@ func NewPU(contextID string, puInfo *policy.PUInfo, timeout time.Duration) (*PUC
...
 }
 
 func (p *PUContext) dnsToACLs(dnsList *policy.DNSRuleList, ipcache map[string]bool) {
 
-	rules := new(policy.IPRuleList)
-	for _, dnsrule := range *dnsList {
+	lookupHost := func(dnsrule *policy.DNSRule) error {
+		var rules policy.IPRuleList
+
 		if ips, err := LookupHost(dnsrule.Name); err == nil {
```
CyrilPeponnet, line 140: Why do we have 

// LookupHost is mapped to the function net.LookupHost
var LookupHost = net.LookupHost

controller/pkg/pucontext/pucontext.go
```diff
@@ -116,44 +116,61 @@ func NewPU(contextID string, puInfo *policy.PUInfo, timeout time.Duration) (*PUC
...
 			zap.L().Warn("Failed to resolve dnsrule", zap.Error(err))
+			return err
 		}
+
+		return nil
+	}
+
+	var errDNSNames policy.DNSRuleList
+
+	for _, dnsrule := range *dnsList {
```
CyrilPeponnet: I think the construct should be done in parallel with a global timeout (including the retry)

For now afaik if I have 10 dnsrules to resolve that fails I will have to wait 10 times the timeout twice with 500ms in between. 

Also it will avoid to kill baby bears by doing a no lint for an err check ;)

controller/pkg/pucontext/pucontext.go
0xmchadha: We do that trick to implement unit tests. net.LookupHost is a system specific function which is not required to execute in unit tests. We map it to our implementation and then use that to verify that our code is indeed making lookup calls to specific hosts.

PR #702, opened 2018-12-06
if proto for external is all, then apply explicit udp and tcp rules also

0xmchadha:
If an external network with "all" in protocol field is specified, then apply tcp and udp rules also in acls.


varks (approved):

dstiliadis:
I don't think that this is needed and it is actually convoluted. Putting protocols as all means that you have to one-directional ACLs. This is normal semantics that iptables and everyone else applies and users are used to them.

PR #721, opened 2019-01-14
Fix nat proxy rules

dstiliadis:
NAT Proxy rules would trap packets from a remote source when the port was matching the protected service. This PR adds a local filter so that the proxy rules will only trap this traffic for local traffic and thus allow traffic from the protected server to the proxy.


ghost:
```
[
  {
    "project": "proxy-rules",
    "component": "enforcerd",
    "pr-id": "999",
    "commit-sha": "470893458908dd11343a1838666eccd9af8c3201"
  },
  {
    "project": "proxy-rules",
    "component": "trireme-lib",
    "pr-id": "721",
    "commit-sha": "d2d3499969572de5ca1e5c6ebb9c71cee2ed3258"
  }
]
```

varks (approved):

PR #728, opened 2019-01-22
Refactor packet library

0xmchadha:
Refactor the packet library so Ipv6 packets can also be accommodated in the enforcer.

### DIFF
--- a/controller/internal/enforcer/nfqdatapath/datapath_tcp.go
+++ b/controller/internal/enforcer/nfqdatapath/datapath_tcp.go
@@ -23,6 +23,11 @@ var errInvalidNetState = errors.New("Invalid net state")
 var errNonPUTraffic = errors.New("Traffic belongs to a PU we are not monitoring")
 var errNetSynNotSeen = errors.New("Network Syn packet was not seen")
 var errNoConnFound = errors.New("no context or connection found")
+var errRejectPacket = errors.New("Reject the packet as per policy")
+var errTCPAuthNotFound = errors.New("TCP authentication option not found")
+var errSigValidFailed = errors.New("Ack packet dropped because signature validation failed")
+var errInvalidFormat = errors.New("Ack packet dropped because of invalid format")
+var errInvalidConnState = errors.New("Invalid connection state")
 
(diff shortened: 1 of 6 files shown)


controller/internal/enforcer/nfqdatapath/datapath_tcp.go
```diff
@@ -800,18 +805,20 @@ func (d *Datapath) processNetworkAckPacket(context *pucontext.PUContext, conn *c
...
 			// d.reportRejectedFlow(tcpPacket, conn, collector.DefaultEndPoint, context.ManagementID(), context, collector.InvalidHeader, nil, nil)
-			return nil, nil, fmt.Errorf("TCP authentication option not found: %s", err)
+			return nil, nil, errTCPAuthNotFound
 		}
 
 		if _, err := d.tokenAccessor.ParseAckToken(&conn.Auth, tcpPacket.ReadTCPData()); err != nil {
 			d.reportRejectedFlow(tcpPacket, conn, collector.DefaultEndPoint, context.ManagementID(), context, collector.InvalidToken, nil, nil)
-			return nil, nil, fmt.Errorf("Ack packet dropped because signature validation failed: %s", err)
+			zap.L().Debug("Error: Ack Packet dropped because signature validation failed", zap.Error("error", err))
+			return nil, nil, errSigValidFailed
```
amitlimaye, line 814: don't use a constant error here. parse token can fail due to multiple reasons. need the error report by that function to propagate

controller/internal/enforcer/nfqdatapath/datapath_tcp.go
```diff
@@ -858,7 +865,7 @@ func (d *Datapath) processNetworkAckPacket(context *pucontext.PUContext, conn *c
 		zap.String("net-conn", hash),
 	)
 
-	return nil, nil, fmt.Errorf("Ack packet dropped, invalid duplicate state: %d", conn.GetState())
+	return nil, nil, errInvalidConnState
```
amitlimaye, line 868: same thing here what state were we in is lost

PR #773, opened 2019-04-16
Introduce a new package which provides LPM lookups and allows methods on the keys and values.

0xmchadha:
#### Description

#### Test plan

> Fixes #.


controller/internal/enforcer/acls/acl.go
```diff
@@ -1,100 +1,106 @@
...
-		plenRules.rules[subnet] = append(plenRules.rules[subnet], r)
-		return nil
-	}
+	return report, packet, err
+}
+
+func (a *acl) addRule(rule policy.IPRule) (err error) {
 
 	for _, proto := range rule.Protocols {
 		if strings.ToLower(proto) == constants.TCPProtoNum {
```
satyamsi: this is a good candidate for - 
if x != y {
    continue
}

the following code should be a separate function.

controller/internal/enforcer/acls/acl_test.go
```diff
@@ -58,7 +58,7 @@ var (
 				PolicyID: "udp0/0"}},
 	}
 	// rulesPrefixLens holds unique prefix lens in rules above.
-	rulesPrefixLens = 5
+//	rulesPrefixLens = 5
```
satyamsi: remove commented out code if really plan to remove this code.

controller/internal/enforcer/applicationproxy/servicecache/servicecache.go
```diff
@@ -207,30 +220,32 @@ func (s *ServiceCache) addHostService(e *common.Service, record *entry, local bo
...
-		entries, ok := prefix[binPrefix]
-		if !ok {
-			continue
-		}
-
-		for _, e := range entries {
-			if e.ports.IsIncluded(port) {
-				return e.data
+	findMatch := func(val interface{}) bool {
+		if val != nil {
```
satyamsi, line 235: same style thing here .. too much indentation within a function.
if val == nil {
    return false
}

utils/ipprefix/ipprefix.go
```diff
@@ -0,0 +1,166 @@
...
+	RunIP(net.IP, func(val interface{}) bool)
+}
+
+const (
+	ipv4MaskSize = 32 + 1
+	ipv6MaskSize = 128 + 1
+)
+
+type ipcache struct {
+	ipv4 []map[uint32]interface{}
```
satyamsi, line 39: locks ?

satyamsi: why is ipv4 and ipv6 both in the same struct - I believe these should be two different types that implement the same interface IPcache. That looks correct mainly because most functions then have an if then else check on v4 vs v6 checks.

satyamsi (changes requested):

controller/internal/enforcer/applicationproxy/servicecache/servicecache_test.go
```diff
@@ -128,20 +128,17 @@ func TestServiceCache(t *testing.T) {
 			cerr := c.Add(s1, "1", "first data", true)
 			So(cerr, ShouldBeNil)
 			So(c.local, ShouldNotBeNil)
-			So(len(c.local), ShouldEqual, 3)
 			So(c.localHosts, ShouldNotBeNil)
```
dstiliadis, line 131: The test is not testing without this. We must read the cache and validate the entries are there in order for the test ot be complete.

utils/ipprefix/ipprefix_test.go
```diff
@@ -0,0 +1,115 @@
...
+				delete(m, s)
+			}
+
+			return val
+		}
+
+		ipcache.RunVal(testRunVal)
+		So(len(m), ShouldEqual, 0)
+	})
+
```
dstiliadis, line 101: Let's add tests for mask=0, and multiple entries in the same mask.

utils/ipprefix/ipprefix.go
dstiliadis: we need locks. I don't see how it can be done without two structures given that the index in the map is the mask. Otherwise the ipv5 vs v6 becomes the callers problem.

utils/ipprefix/ipprefix.go
```diff
@@ -0,0 +1,166 @@
...
+		ip = ip.To4()
+
+		if cache.ipv4[mask] == nil {
+			cache.ipv4[mask] = map[uint32]interface{}{}
+		}
+
+		m := cache.ipv4[mask]
+		// the following expression is ANDing the ip with the mask
+		m[binary.BigEndian.Uint32(ip)&binary.BigEndian.Uint32(net.CIDRMask(mask, 32))] = val
+	} else {
```
dstiliadis: return and avoid the else. I suspect the linter will complain as well. Same in the below methods.

utils/ipprefix/ipprefix.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+
+			m[ip] = v
+		}
+
+		if len(m) == 0 {
+			cache.ipv6[mask] = nil
+		}
+	}
+
```
dstiliadis, line 233: Running this in both ipv4 and ipv6 might create problems for the callers. Something doesn't feel right here.

controller/internal/enforcer/acls/acl_test.go
```diff
@@ -180,8 +180,8 @@ func TestObservedLookup(t *testing.T) {
 			err := a.addRule(r)
 			So(err, ShouldBeNil)
 		}
-		a.reverseSort()
-		So(len(a.prefixLenMap), ShouldEqual, rulesWithObservationPrefixLens)
+		//a.reverseSort()
+		///So(len(a.prefixLenMap), ShouldEqual, rulesWithObservationPrefixLens)
```
dstiliadis, line 190: let's read the cache and keep the validations in the tests. Just in case.

controller/internal/enforcer/applicationproxy/servicecache/servicecache_test.go
0xmchadha: I agree, but this particular test is just testing whether the add entries succeeds or not. There are following tests which are indeed doing a cache lookup after insertions.

controller/internal/enforcer/acls/acl_test.go
0xmchadha: Added the cache lookups after insertion here.

utils/ipprefix/ipprefix.go
0xmchadha: Added locks in the new commit.

utils/ipprefix/ipprefix.go
0xmchadha: Over here we want to process all the values stored in the cache independent of the keys. If the client stores ipv4 and ipv6 addresses in a combined cache and it wants to process all the values, then this method is used

utils/ipprefix/ipprefix_test.go
0xmchadha: Yes, that should be there. My bad. I will add the test with mask = 0.

dstiliadis (dismissed):

utils/ipprefix/ipprefix.go
```diff
@@ -0,0 +1,237 @@
+package ipprefix
+
+import (
+	"encoding/binary"
+	"net"
+	"sync"
+)
+
+// FuncOnKeys is the type of func which will operate on the value associated with the lpm ip.
+type FuncOnKeys func(val interface{}) bool
```
satyamsi: consistency: Based on your naming, this should be called FuncOnLpmIP

utils/ipprefix/ipprefix.go
```diff
@@ -0,0 +1,237 @@
...
+// associated value.
+type FuncOnVals func(val interface{}) interface{}
+
+//IPcache is an interface which provides functionality to store ip's and do longest prefix match
+type IPcache interface {
+	// Put takes an argument an ip address, mask and value.
+	Put(net.IP, int, interface{})
+	// Get takes an argument the IP address and mask and returns the value that is stored for that key.
+	Get(net.IP, int) (interface{}, bool)
+	// RunFuncOnLpmIP function takes as an argument an IP address and a function. It finds the subnet to which this IP belongs with the longest prefix match.
```
satyamsi: consistency: Line wrap to match other descriptions. I dont have a strong opinion on how wide - just consistent.

utils/ipprefix/ipprefix.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (cache *ipcacheV4) Get(ip net.IP, mask int) (interface{}, bool) {
+	cache.RLock()
+	defer cache.RUnlock()
+
+	m := cache.ipv4[mask]
+	if m != nil {
+		val, ok := m[binary.BigEndian.Uint32(ip)&binary.BigEndian.Uint32(net.CIDRMask(mask, 32))]
+
```
satyamsi: code-blocks: delete extra lline. typically a code block

```
val, err := DoSomething()
if err != nil {
    return err
}
```

as opposed to 

```
val, err := DoSomething()

if err != nil {
    return err
}
```

utils/ipprefix/ipprefix.go
```diff
@@ -0,0 +1,237 @@
...
+	copy(maskip[:], ip.Mask(net.CIDRMask(mask, 128)))
+	m[maskip] = val
+}
+
+func (cache *ipcacheV6) Get(ip net.IP, mask int) (interface{}, bool) {
+	cache.RLock()
+	defer cache.RUnlock()
+
+	m := cache.ipv6[mask]
+
```
satyamsi: code-blocks: remove extra line.

utils/ipprefix/ipprefix.go
```diff
@@ -0,0 +1,237 @@
...
+	if m != nil {
+		var maskip [16]byte
+		copy(maskip[:], ip.Mask(net.CIDRMask(mask, 128)))
+		val, ok := m[maskip]
+
```
satyamsi: code-blocks: remove extra line.

controller/internal/enforcer/acls/acl_test.go
```diff
@@ -71,8 +69,6 @@ func TestLookup(t *testing.T) {
 			err := a.addRule(r)
```
satyamsi, line 69: why are we using a global slice of rules - if this is the only test lets bring it in here.

satyamsi (dismissed):

utils/ipprefix/ipprefix.go
0xmchadha: Fixed.

utils/ipprefix/ipprefix.go
0xmchadha: Fixed.

utils/ipprefix/ipprefix.go
0xmchadha: Fixed.

controller/internal/enforcer/acls/acl_test.go
0xmchadha: fixed.

utils/ipprefix/ipprefix.go
0xmchadha: Fixed.

utils/ipprefix/ipprefix.go
0xmchadha: Fixed.

amitlimaye (dismissed):

dstiliadis (dismissed):

amitlimaye (approved):

PR #787, opened 2019-05-08
Fix logging bug in nfq_linux

0xmchadha:
This PR fixes the following issues:

1) First if the packet creation fails the logging later will panic as it accesses the packet field.
2) The actual reason of the packet drop from the above layers is not logged.
3) Add debug Log in the application path also if the packet is dropped.

### DIFF
--- a/controller/internal/enforcer/nfqdatapath/nfq_linux.go
+++ b/controller/internal/enforcer/nfqdatapath/nfq_linux.go
@@ -76,23 +76,26 @@ func (d *Datapath) startApplicationInterceptor(ctx context.Context) {
 // processNetworkPacketsFromNFQ processes packets arriving from the network in an NF queue
 func (d *Datapath) processNetworkPacketsFromNFQ(p *nfqueue.NFPacket) {
 
-	// Parse the packet - drop if parsing fails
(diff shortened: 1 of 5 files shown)


controller/internal/enforcer/nfqdatapath/nfq_linux.go
```diff
@@ -103,14 +107,14 @@ func (d *Datapath) processNetworkPacketsFromNFQ(p *nfqueue.NFPacket) {
 		)
 		length := uint32(len(p.Buffer))
 		buffer := p.Buffer
-		p.QueueHandle.SetVerdict2(uint32(p.QueueHandle.QueueNum), 0, uint32(p.Mark), length, uint32(p.ID), buffer)
+		p.QueueHandle.SetVerdict2(0, uint32(p.Mark), length, uint32(p.ID), buffer)
 		if netPacket.IPProto() == packet.IPProtocolTCP {
```
dstiliadis, line 112: why 0 ? doesn't look right

0xmchadha: I removed the first argument. QueueNum is not required in SetVerdict function.

PR #803, opened 2019-05-23
New Kubernetes Monitor - Phase 2

mheese:
Phase 2 of the new kubernetes monitor. Addresses the following:

- Fixes https://github.com/aporeto-inc/aporeto/issues/1209


monitor/internal/pod/controller.go
```diff
@@ -54,6 +54,7 @@ func newReconciler(mgr manager.Manager, handler *config.ProcessorConfig, metadat
 
 		// TODO: might move into configuration
 		handlePUEventTimeout:   5 * time.Second,
+		getterTimeout:          2 * time.Second,
```
dstiliadis, line 57: The above is very short as discussed.

mheese:
integration tests failed with the error that Sibi is working on right now, as it passed FT, I'm going to merge now anyway.

monitor/internal/pod/controller.go
mheese: and handled through reconciling on errors anyway which is a must at the moment as explained. will change in the phase 3 of the implementation then.

PR #822, opened 2019-06-25
Envoyproxy POC

mheese:


controller/internal/enforcer/envoyproxy/authz/server.go
```diff
@@ -0,0 +1,501 @@
...
+			if httpReq != nil {
+				httpReqHeaders := httpReq.GetHeaders()
+				if httpReqHeaders != nil {
+					aporetoKey, _ = httpReqHeaders[aporetoKeyHeader]
+					aporetoAuth, _ = httpReqHeaders[aporetoAuthHeader]
+				}
+			}
+		}
+	}
+
```
dstiliadis, line 223: if attrs is nil it flows here. Is that what is needed. Shouldn't we just return for nil attributes? The httpReq might also be nil and it is used below.

controller/internal/enforcer/envoyproxy/authz/server.go
```diff
@@ -0,0 +1,501 @@
...
+	// now we can see if we can decode our claims
+	srcid, scopes, profile, err := s.verifier.ParseToken(aporetoAuth, aporetoKey)
+	if err != nil {
+		zap.L().Debug("ext_authz ingress: failed to parse Aporeto token", zap.String("puID", s.puID), zap.Error(err))
+		flow.DropReason = "failed to parse Aporeto token"
+		return createDeniedCheckResponse(rpc.PERMISSION_DENIED, envoy_type.StatusCode_Forbidden, "failed to parse Aporeto token"), nil
+	}
+	zap.L().Debug("ext_authz ingress: parsed Aporeto token", zap.String("puID", s.puID), zap.String("srcid", srcid), zap.Strings("scopes", scopes), zap.Strings("profile", profile))
+	flow.Source.ID = srcid
+
```
dstiliadis, line 246: I guess there is no API authorization yet .. To be added. I would do the network checks before the API authorization logic. No need to spend time decoding tokens if the network policy rejects it. I think these are backwards.

controller/internal/enforcer/envoyproxy/authz/server.go
```diff
@@ -0,0 +1,501 @@
...
+		zap.L().Error("ext_authz ingress: failed to get PU context", zap.String("puID", s.puID), zap.Error(err))
+		return createDeniedCheckResponse(rpc.INTERNAL, envoy_type.StatusCode_InternalServerError, "failed to get PU context"), nil
+	}
+	pctx, ok := pctxRaw.(*pucontext.PUContext)
+	if !ok {
+		zap.L().Error("ext_authz ingress: PU context has the wrong type", zap.String("puID", s.puID), zap.String("puContextType", fmt.Sprintf("%T", pctxRaw)))
+		return createDeniedCheckResponse(rpc.INTERNAL, envoy_type.StatusCode_InternalServerError, "PU context has the wrong type"), nil
+	}
+
+	// extract our headers
```
dstiliadis, line 195: here you need the sctx of the app proxy. There can be multiple services listening for a given PU. I am not sure if we can recover the original destination/port from the request or how envoy will provide this.

controller/internal/enforcer/envoyproxy/authz/server.go
```diff
@@ -0,0 +1,501 @@
...
+	}, nil
+}
+
+// egressCheck implements the AuthorizationServer for egress connections
+func (s *Server) egressCheck(ctx context.Context, checkRequest *ext_authz_v2.CheckRequest) (*ext_authz_v2.CheckResponse, error) {
+	zap.L().Debug("ext_authz egress: checkRequest", zap.String("puID", s.puID), zap.String("checkRequest", checkRequest.String()))
+
+	// get the PU context
+	pctxRaw, err := s.puContexts.Get(s.puID)
+	if err != nil {
```
dstiliadis, line 277: same here. We need one more local to discover the destination service context, from there the API specification, and from there the decision is made on whether this is external or internal, public or not, and so on. See here for the parallel: 

https://github.com/aporeto-inc/trireme-lib/blob/e08d901036cda50444be7f4285d41103d65397a6/controller/internal/enforcer/applicationproxy/http/http.go#L367

The question is whether the original destination/port are known on the envoy API.

Services are indexed by original destination. For every address/port pair there can be a different service that is being targeted.

controller/internal/enforcer/envoyproxy/authz/server.go
```diff
@@ -0,0 +1,501 @@
...
+	defer unlockSecrets()
+	transmittedKey := secrets.TransmittedKey()
+	token, err := servicetokens.CreateAndSign(
+		sourceAdress,
+		pctx.Identity().Tags,
+		pctx.Scopes(),
+		pctx.ManagementID(),
+		defaultValidity,
+		secrets.EncodingKey(),
+	)
```
dstiliadis, line 334: Just FYI: this will go away with the OAuth PRs .. we will stop transmitting the signed public key and ask the servers to hit the OAUTH JWKS API to retrieve it from the token issuer. This is coming. Just noting it here at the time being.

controller/internal/enforcer/envoyproxy/authz/server.go
dstiliadis: We obviously need to add the user claims from the Authorization Bearer header. 

I am noticing lots of commonality with the application proxy. I think we should refactor the app proxy, and use it as a library in both places and then we don't repeat the logic twice. The processAppRequest, processNetRequest functions are pretty similar with minor modifications on what they return. If we refactor the logic out, we can use a common code based whether we are inline or not.

mheese: I started noticing that as well, and it definitely would have been one of my proposals: we need to make parts of the app proxy reusable in places like this.

re the sctx: I'm not sure how there could be any service listening at all for a PU here: this currently does not support at all the running application proxy, nor is it started for this enforcer.

The biggest trouble here for us is that we actually would need to distinguish between Istio pods, and just plain envoy pods where we actually might want to setup iptables ourselves, and run an app proxy. In Istio this is completely impossible because they take over all the traffic.

For this enforcer in general I think we should go about it more the "developer" way: make things configurable and make the user partially responsible for making it work. This provides flexibility which is exactly what people would like to have when they choose this enforcer. If they don't want that, well just go with our standard enforcer.

controller/internal/enforcer/envoyproxy/authz/server.go
mheese: if attrs is nil, it will not have extracted the headers and we will fail a couple of lines below

controller/internal/enforcer/envoyproxy/authz/server.go
mheese: the destination IP and port contains the IP and port as used outgoing by Envoy. That means: pod IP + the port as specified in the pod template (yes, Istio _requires_ you to fill out that part of your yamls)

controller/internal/enforcer/envoyproxy/authz/server.go
mheese: cool :) ... looking forward to that!

PR #824, opened 2019-06-27
Changing native IDs to k8s UIDs

mheese:


monitor/internal/pod/delete_controller.go
```diff
@@ -0,0 +1,132 @@
...
+			zap.L().Warn("failed to get pod from Kubernetes API", zap.String("puID", nativeID), zap.String("namespacedName", req.String()), zap.Error(err))
+		}
+		return
+	}
+
+	// the edge case: a pod with the same namespaced name came up and we have missed a delete event
+	// this means that this pod belongs to a different PU and must live, therefore we try to delete the old one
+	if nativeID != string(pod.GetUID()) {
+		zap.L().Error("Pod does not have expected native ID, we must have missed an event and the same pod was recreated. Trying to destroy PU", zap.String("puID", nativeID), zap.String("namespacedName", req.String()), zap.String("podUID", string(pod.GetUID())))
+		if err := pc.Policy.HandlePUEvent(
```
abhijitherekar, line 119: where is the creation of PU happening with this new PU-id?

abhijitherekar: ok, I got it.

monitor/internal/pod/delete_controller.go
```diff
@@ -0,0 +1,132 @@
...
+	m := make(map[string]client.ObjectKey)
+
+	// the poor man's controller loop
+	for {
+		select {
+		case ev := <-c.deleteCh:
+			m[ev.NativeID] = ev.Key
+		case <-c.reconcileCh:
+			c.reconcileFunc(backgroundCtx, c.client, c.handler, c.itemProcessTimeout, &m)
+		case <-t.C:
```
abhijitherekar, line 72: So if we miss a delete event then the timer will kick and check with the API server to see if we have the obj, if not delete it, right?
Maybe in the next controller-runtime we can use the direct API-server client while fetching the objects.

abhijitherekar (commented):
As far as I can understand the issue, the changes look good to me.

monitor/internal/pod/controller.go
```diff
@@ -116,38 +128,38 @@ type ReconcilePod struct {
...
-			// call destroy - the policy engine is going to call stop first for us if necessary
-			if err := r.handler.Policy.HandlePUEvent(
-				handlePUCtx,
-				puID,
-				common.EventDestroy,
-				policy.NewPURuntimeWithDefaults(),
-			); err != nil {
-				zap.L().Error("failed to handle destroy event", zap.String("puID", puID), zap.Error(err))
-			}
+			r.deleteReconcileCh <- struct{}{}
```
abhijitherekar, line 137: Just curious,I see you send a event to delete the PU,  I see no harm  in the below logic to be removed. Was there any specific reason?

abhijitherekar (approved):

PR #840, opened 2019-07-30
Fix Kubernetes Monitor Deletes

mheese:
As @dstiliadis and I have debugged this morning, it looks like we can even miss a terminating event, which will mean that the pod will never register with the delete controller, and therefore we will never destroy the PU.

This change simply changes the logic of registering the deletes: whenever we reconcile past the retrieval of the pod object, we will register it with the delete controller, and let it handle it from there.


abhijitherekar (approved):
Makes sense. LGTM.

dstiliadis (approved):

PR #858, opened 2019-08-14
Adding envoy apis

mheese:
Adds a script for generating the envoy APIs from their github repo, and the resulting files.

Implements https://github.com/aporeto-inc/aporeto/issues/1701


mheese:
@dstiliadis @abhijitherekar please look over ... I created a quick test file (not in the PR) which creates an ext_authz and SDS server and they compiled (which means that the generation should have been successful)

If you are okay, I'll just merge this without any tests ... no code touches this yet anyway

abhijitherekar (approved):

PR #863, opened 2019-08-29
Envoy Enforcer plumbing

mheese:


controller/internal/enforcer/nfqdatapath/datapath.go
```diff
@@ -725,7 +725,7 @@ func (d *Datapath) EnableDatapathPacketTracing(contextID string, direction packe
 		d.packetTracingCache.Remove(contextID) // nolint
 	}()
 
-	return nil
+	return ctx.Err()
```
dstiliadis: Why?

dstiliadis (dismissed):

controller/internal/enforcer/nfqdatapath/datapath.go
mheese: let me change that ... it's not really helping/doing anything anymore at that return

dstiliadis (dismissed):

dstiliadis (approved):

mheese:
codecov/path is not going to pass on purpose because the whole implementation is still up to be done.

All changes have their respective unit tests properly adjusted.

Merging now.

PR #909, opened 2019-10-19
Clean caches on RST

dstiliadis:
This PR cleans up the caches on early RST packets. In some installations we are noticing that a Syn/SynAck sequence is followed by an immediate RST. When this happens the kernel re-uses immediately the source port and this creates state issues in the connection management (unlike normal cases where the ports are rotated over a large number of connections).

If one of the initial negotiation packets is RST there is no reason to keep state. We let the packet go and immediately clean up the cache. 

The problem was first identified as part of the OpenShift router that seems to be doing keep-alive by initiating a Syn/SynAck sequence followed by an immediate RST. This leads the kernel in immediately re-using the source port and this confuses the state machines in the case where host containers are protected.

### DIFF
--- a/controller/internal/enforcer/nfqdatapath/datapath_tcp.go
+++ b/controller/internal/enforcer/nfqdatapath/datapath_tcp.go
@@ -61,15 +61,17 @@ func (d *Datapath) processNetworkTCPPackets(p *packet.Packet) (conn *connection.
 
 	case packet.TCPSynAckMask:
 		conn, err = d.netSynAckRetrieveState(p)
-		switch err {
-		case pucontext.ToError(pucontext.ErrOutOfOrderSynAck):
-			// Drop this synack it is for a flow we know which is marked for deletion.
-			// We saw a FINACK and this synack has come without we seeing an appsyn for this flow again
-			//return conn, fmt.Errorf("%s %s:%d", err, p.SourceAddress().String(), int(p.SourcePort()))
-			return conn, pucontext.PuContextError(pucontext.ErrOutOfOrderSynAck, fmt.Sprintf("%s %s:%d", err, p.SourceAddress().String(), int(p.SourcePort())))
-		case pucontext.ToError(pucontext.ErrNonPUTraffic):
-			d.releaseUnmonitoredFlow(p)
-			return conn, nil
+		if err != nil {
+			switch err {
+			case pucontext.ToError(pucontext.ErrOutOfOrderSynAck):
+				// Drop this synack it is for a flow we know which is marked for deletion.
+				// We saw a FINACK and this synack has come without we seeing an appsyn for this flow again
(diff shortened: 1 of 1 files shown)


dstiliadis:
```
[
  {
    "project": "",
    "component": "enforcerd",
    "pr-id": "1471",
    "commit-sha": "901b317a2048314d234474ad3ac353e5d51b1586"
  },
  {
    "project": "",
    "component": "trireme-lib",
    "pr-id": "909",
    "commit-sha": "6770310e58faf507c5f8b258032e32af2c0ef894"
  }
]
```

controller/internal/enforcer/nfqdatapath/datapath_tcp.go
```diff
@@ -61,15 +61,17 @@ func (d *Datapath) processNetworkTCPPackets(p *packet.Packet) (conn *connection.
...
-			d.releaseUnmonitoredFlow(p)
-			return conn, nil
+		if err != nil {
+			switch err {
+			case pucontext.ToError(pucontext.ErrOutOfOrderSynAck):
+				// Drop this synack it is for a flow we know which is marked for deletion.
+				// We saw a FINACK and this synack has come without we seeing an appsyn for this flow again
+				//return conn, fmt.Errorf("%s %s:%d", err, p.SourceAddress().String(), int(p.SourcePort()))
+				return conn, pucontext.PuContextError(pucontext.ErrOutOfOrderSynAck, fmt.Sprintf("%s %s:%d", err, p.SourceAddress().String(), int(p.SourcePort())))
+			default:
```
satyamsi, line 71: instead of adding explicitly a new error for RST, you have used default. what i cant tell is if we have started doing this for other unwanted errors. maybe ok.

controller/internal/enforcer/nfqdatapath/datapath_tcp.go
```diff
@@ -132,6 +134,17 @@ func (d *Datapath) processNetworkTCPPackets(p *packet.Packet) (conn *connection.
 
 	}
 
+	if p.GetTCPFlags()&packet.TCPRstMask != 0 {
```
satyamsi, line 137: i'm a little concerned about this. i have seen RSTs flying back and forth at the end of connections interleaved with FINs. This I believe needs more testing in other cases.

satyamsi (approved):

controller/internal/enforcer/nfqdatapath/datapath_tcp.go
dstiliadis: there is nothing else that the function can return.

controller/internal/enforcer/nfqdatapath/datapath_tcp.go
dstiliadis: This only catches RSTs during the 3-way handshake.

PR #943, opened 2020-01-14
Master for 3.13

mheese:
Merging master into 3.13 for release.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -14,41 +14,32 @@ required = ["github.com/docker/distribution"]
 
 [[constraint]]
   name = "github.com/ti-mo/conntrack"
-  revision = "309d2d6a30782ae8f0c68f2dda0774966523e469"
-  source = "https://github.com/aporeto-inc/conntrack.git"
-
-[[constraint]]
-  name = "k8s.io/api"
-  version = "kubernetes-1.13.1"
-
-[[constraint]]
-  name = "k8s.io/client-go"
-  version = "kubernetes-1.13.1"
-
-[[constraint]]
-  branch = "release-3.13.0"
-  name = "github.com/aporeto-inc/oxy"
-
-[[constraint]]
-  branch = "release-3.13.0"
-  name = "go.aporeto.io/netlink-go"
-
-[[constraint]]
-  name = "github.com/hashicorp/go-version"
-  version = "v1.0.0"
+  branch = "master"
 
 [[override]]
   name = "github.com/ti-mo/netfilter"
-  version = "<0.3.0"
+  version = "=0.3.0"
 
-[[override]]
+# as per github.com/ti-mo/netfilter go.mod file
+[[constraint]]
   name = "github.com/mdlayher/netlink"
-  revision = "7e948d4190d773b1315fb031a8ae83d8a302f7c9"
-  source = "https://github.com/aporeto-inc/netlink.git"
+  revision = "a1644773bc99cfd147417f646434a663b5892505"
 
+#
+# The most significant dependency for the Kubernetes monitor: the controller-runtime
+# NOTE: change with care and always adjust the Kubernetes dependencies below
+#
 [[override]]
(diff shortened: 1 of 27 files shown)


Gopkg.toml
```diff
@@ -14,42 +14,33 @@ required = ["github.com/docker/distribution"]
...
-  name = "github.com/aporeto-inc/oxy"
-
-[[constraint]]
-  branch = "release-3.13.0"
-  name = "go.aporeto.io/netlink-go"
-
-[[constraint]]
-  name = "github.com/hashicorp/go-version"
-  version = "v1.0.0"
+  branch = "master"
```
brianonn, line 17: @mheese my only thought is : do we want github.com/ti-mo/conntrack:**master** ?
I think it might be better to pin it to the current master revision @ sha c9b176489c1a057922cbaa0a78df9e51cffbcc0d

```
  name = "github.com/ti-mo/conntrack"
  revision = "c9b176489c1a057922cbaa0a78df9e51cffbcc0d"
  source = "https://github.com/ti-mo/conntrack.git"
```

controller/internal/enforcer/applicationproxy/http/http.go
```diff
@@ -182,6 +186,30 @@ func (p *Config) GetClientCertificateFunc(_ *tls.CertificateRequestInfo) (*tls.C
 	p.RLock()
 	defer p.RUnlock()
 	if p.cert != nil {
+		cert, err := x509.ParseCertificate(p.cert.Certificate[0])
+		if err != nil {
+			zap.L().Error("http: Cannot build the cert chain")
```
brianonn, line 191: I'd like to make this error different from the next one @ L197 -- if it shows up in a log we can quickly see where it came from. 
"http: Cannot parse the certificate"

controller/internal/enforcer/envoyauthorizer/envoyauthorizerenforcer.go
```diff
@@ -82,16 +97,195 @@ func (e *Enforcer) Secrets() (secrets.Secrets, func()) {
...
+		egressServer, err := envoyproxy.NewExtAuthzServer(contextID, e.puContexts, e.collector, envoyproxy.EgressDirection, e.registry, e.secrets, e.tokenIssuer)
+		if err != nil {
+			zap.L().Error("Cannot create and run EgressServer", zap.Error(err))
+			ingressServer.Stop()
+			return err
+		}
+		sdsServer, err := envoyproxy.NewSdsServer(contextID, puInfo, caPool, e.secrets)
+		if err != nil {
+			zap.L().Error("Cannot create and run SdsServer", zap.Error(err))
+			return err
```
brianonn, line 146: Do you need to stop the ingressServer and egressServer before returning ?
```
ingressServer.Stop()
egressServer.Stop()
return err
```

brianonn (commented):
comments left

controller/internal/enforcer/envoyauthorizer/envoyauthorizerenforcer.go
mheese: looks correct yes

Gopkg.toml
mheese: there is not even a constraint in general for this anymore, that's why I just put this to master

controller/internal/enforcer/applicationproxy/http/http.go
mheese: that would indeed be better

controller/internal/enforcer/envoyauthorizer/envoyauthorizerenforcer.go
abhijitherekar: My Bad, good catch @brianonn  , will make the changes and commit to master.

mheese:
merging this now ... we don't have time for nit picking ... the purpose of a review here was to ensure all changes from master that we need go in ... and they are, everything else needs to be handled separately

PR #945, opened 2020-01-15
Follow-up from #943

mheese:
Follow-up from #943 . Adding better logging for http client certs and adding stops in envoy.

There could be a lot more done in both packages that are touched here. However, especially the http package stands out: logging is not very clear, UT missing (in the works though for envoy by Abhi), a lot of variable/pointer checks missing, unclear return values and missing errors - should at least have a comment if they are intended, ignoring golang best practices and not very elegant code. One can see that this was written down in a hurry, and fixed in a hurry by us recently as well :) This package deserves some love.

Anyway, I leave it up to @brianonn and @abhijitherekar to add more to this PR if you think that is necessary.

@brianonn I ignored the `Gopkg.toml` comment from #943 as I don't think we need to change anything. Please comment here if you think otherwise.

### DIFF
--- a/controller/internal/enforcer/applicationproxy/http/http.go
+++ b/controller/internal/enforcer/applicationproxy/http/http.go
@@ -185,16 +185,16 @@ func (p *Config) newBaseTLSClientConfig() *tls.Config {
 func (p *Config) GetClientCertificateFunc(_ *tls.CertificateRequestInfo) (*tls.Certificate, error) {
 	p.RLock()
 	defer p.RUnlock()
-	if p.cert != nil {
+	if p.cert != nil && len(p.cert.Certificate) > 0 {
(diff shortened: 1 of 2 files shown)


brianonn (commented):
Nice! thanks @mheese . I like the way you made your changes. 

I still think we should pin the Gopkg.toml entry for conntrack and not let it float around with master. 
```
[[constraint]]
  name = "github.com/ti-mo/conntrack"
  revision = "c9b176489c1a057922cbaa0a78df9e51cffbcc0d"
```
Is there a reason why you think this is not necessary?  I never like choosing a branch tip and I always like to have it pinned to a commit and reproducible, rather than allow the version we pull in to change (and possibly open it to new bugs or security holes that are introduced)

PR #951, opened 2020-01-28
fix: panic when the token is malformed

sibicramesh:
**NOTE: This fix only avoids the crash. This doesn't explain why the token was messed up in the first place.**

The issue here is the signature position in the token is missing and the check for that was absent. We then proceed to extract the signature from this position in this case will be `0`. So is this panic
```panic
panic: runtime error: slice bounds out of range [22:0]
```
Added unit tests to avoid regression.

Full stack here 
[enforcerd.log](https://github.com/aporeto-inc/trireme-lib/files/4125107/enforcerd.log)

### DIFF
--- a/controller/pkg/tokens/binaryjwt.go
+++ b/controller/pkg/tokens/binaryjwt.go
@@ -516,7 +516,7 @@ func unpackToken(isAck bool, data []byte) ([]byte, []byte, []byte, []byte, error
 	sigPosition := int(binary.BigEndian.Uint16(data[lengthPosition : lengthPosition+2]))
 
 	// The token must be long enough to have at least 1 byte of signature.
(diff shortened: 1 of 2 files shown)


sibicramesh:
```
[
  {
    "project": "",
    "component": "enforcerd",
    "pr-id": "1551",
    "commit-sha": "dc8842341f71ce64692c8184b81a3a97bca94403",
    "pipeline": "master"
  },
  {
    "project": "",
    "component": "trireme-lib",
    "pr-id": "951",
    "commit-sha": "9cbbc4d1b9c17c5ba77bca07c0b623d79c09e536",
    "pipeline": "master"
  }
]
```

amitlimaye (approved):

PR #977, opened 2020-02-25
Windows diagnostics

tillery-aporeto:
Implemented the diagnostics ping feature for Windows.

### DIFF
--- a/controller/internal/enforcer/nfqdatapath/diagnostics_tcp_windows.go
+++ b/controller/internal/enforcer/nfqdatapath/diagnostics_tcp_windows.go
@@ -4,7 +4,6 @@ package nfqdatapath
 
 import (
 	"fmt"
-	"math/rand"
 	"net"
 	"syscall"
 	"unsafe"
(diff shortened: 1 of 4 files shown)


controller/internal/enforcer/nfqdatapath/diagnostics_tcp_windows.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func createConnection(srcIP, dstIP net.IP) (*diagnosticsConnection, error) {
+	return &diagnosticsConnection{}, nil
+}
+
+func (diagConn *diagnosticsConnection) Close() {
+}
+
+func IPtoUint32Array(ip net.IP) [4]uint32 {
```
philipatl: I would make this private, and put some kind of comment about what it's doing (putting the ip address in a format that the driver expects).

tillery-aporeto: Will do.

controller/internal/enforcer/nfqdatapath/diagnostics_tcp_windows.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// payload.
+	payload := raw.Make()
+	payload.Data = tcpData
+
+	tcpPacket.SetPayload(payload)  // nolint:errcheck
+	ipPacket.SetPayload(tcpPacket) // nolint:errcheck
+
+	// pack the layers together.
+	buf, err := layers.Pack(ipPacket, tcpPacket, payload)
```
philipatl, line 92: Can some of this function's code be made common and called by both windows and non-windows impl? And side note: why is ipPacket included here and not for linux?

tillery-aporeto: The Windows Implementation needs the ipPacket because it needs the wire packet to including the IP header and Linux doesn't.  I moved createPacket function back to diagnostics_tcp.go and created more a more specific function called createWirePacket, and now there is more common code.

philipatl (approved):

PR #981, opened 2020-03-04
Changing Kubernetes monitor startup behaviour

mheese:
The BCGov has seen fatal errors during the enforcer startup like this:

```
{"l":"fatal","t":1583278316.241352,"c":"run/run.go:570","m":"Unable to start monitors","error":"pod: controller did not start within 5s"}
```

This will change the behaviour to just log a warning and waiting for the controllers to start. This will block indefinitely until the controllers have started.

Addresses https://github.com/aporeto-inc/aporeto/issues/2692


monitor/internal/pod/monitor.go
```diff
@@ -183,8 +183,8 @@ func (m *PodMonitor) Run(ctx context.Context) error {
 	case err := <-errCh:
 		return fmt.Errorf("pod: %s", err.Error())
 	case <-time.After(5 * time.Second):
-		// we give the controller 5 seconds to report back
-		return errors.New("pod: controller did not start within 5s")
+		// we give the controller 5 seconds to report back before we issue a warning
```
abhijitherekar: Can we also log how many seconds did it take to start the controller.

abhijitherekar: Also, if we just return from here, there will be a mis-leading log saying enforcerd is ready in the run.go, where as we are still retrying.  So, we can make this log more verbose. ?

abhijitherekar: maybe this help us in future debugging.

mheese: good idea ... I'll add that as a debug log.

monitor/internal/pod/monitor.go
```diff
@@ -179,24 +178,28 @@ func (m *PodMonitor) Run(ctx context.Context) error {
...
-	case <-controllerStarted:
-		m.kubeClient = mgr.GetClient()
-
-		// call ResyncWithAllPods before we return from here
-		// this will block until every pod at this point in time has been seeing at least one `Reconcile` call
-		// we do this so that we build up our internal PU cache in the policy engine,
-		// so that when we remove stale pods on startup, we don't remove them and create them again
-		if err := ResyncWithAllPods(ctx, m.kubeClient, m.resyncInfo, m.eventsCh, m.localNode); err != nil {
-			zap.L().Warn("Pod resync failed", zap.Error(err))
+waitLoop:
```
abhijitherekar, line 224: do we want to make this a goroutine, so that we don't block the monitor run?
Or does it conflict with the ResyncWithAllPods logic? As I remember there is a reason we wanted this to be blocking, am I right?

monitor/internal/pod/monitor.go
mheese: looks like this now:

```
DEBUG	pod: controller startup finished	{"duration": "100.176139ms"}
```

monitor/internal/pod/monitor.go
mheese: No, the whole point here is to block the monitor run. And yes, it is related to the resync. We must have completed the resync once before we can release the monitor.

monitor/internal/pod/monitor.go
mheese: yes, I agree

mheese:
```
[
  {
    "project": "k8s-startup",
    "component": "enforcerd",
    "pr-id": "1601",
    "commit-sha": "5bc912993451feb292b1ff0daa01878911ac68b4",
    "pipeline": "master"
  },
  {
    "project": "k8s-startup",
    "component": "trireme-lib",
    "pr-id": "981",
    "commit-sha": "18ac0a8851bc87c7669b1004a23d6e21d220a5cd",
    "pipeline": "master"
  }
]
```

abhijitherekar (approved):

amitlimaye (approved):

mheese:
going to merge this now. It turns out that all the previous FT runs have been failing because of this issue here: https://ci.aporeto.io/teams/main/pipelines/pr-tests/jobs/functional-tests/builds/2932

As this is unrelated to the change, I'm going to merge this now.

PR #987, opened 2020-03-12
Nfqueuebalance

amitlimaye:
#### Description
move from nfqueuebalance to our own hash for loadblancing
#### Test plan

> Fixes #.


utils/cgnetcls/internal/indexallocator/indexallocator.go
```diff
@@ -0,0 +1,62 @@
...
+			startIndex:     startIndex,
+			availableMarks: make(map[int]struct{}, size),
+		}
+		for i := startIndex; i < startIndex+size; i++ {
+			indexes.availableMarks[i] = struct{}{}
+		}
+	})
+	return indexes, size, startIndex
+}
+
```
dstiliadis, line 41: I think it's much easier to do this with a channel? And we already have an allocator doing the same thing. So I am confused as to why we are doing this again?

PR #1050, opened 2021-01-22
Making the DeleteController more robust

mheese:
- improving logging in the delete controller at some places
- swapping controller-runtime client with vanilla Kubernetes golang client
- adding retry logic before we send the delete event


abhijitherekar (approved):
LGTM.

